PR #1729, opened 2013-08-06
Only spin .icon-spin:before instead of .icon-spin itself.

tomdeboer:
Same visual result, but only the icon will be spinning, not the whole element.

Giving you this:
screen shot 2013-08-07 at 01 05 52

Rather than this:
screen shot 2013-08-07 at 01 06 30


robmadole:

– FA team

PR #2232, opened 2013-10-24
Chromium bug work-around.

taystack:
This fixed our issue with chromium not showing the proper icons since chromium only accepts tiffs.


robmadole:

– FA team

PR #2323, opened 2013-11-02
Rewrote SCSS to utilize placeholders & make class output optional

mstduff:
Several updates made to SCSS, mainly for the ability to turn off class output to use placeholder selectors for more semantic HTML
## Updates
- added placeholders selectors for every class (http://sass-lang.com/documentation/file.SASS_REFERENCE.html#placeholder_selectors_)
- added variable to decide if classes should be outputted ($fa-show-classes)
- made icon declarations in a list for looping reasons
- made a alias list as well
- made _icons.scss utilize loops of icon and alias lists for management ease
## Usage

``` scss
//using placeholder selector
.my-class {
  @extend %fa;
  &:before {
    @extend %fa-star;
  }
}
// same as: <i class="fa fa-star"></i>

//using with text inline
.my-button {
  &:before {
    @extend %fa-star;
    @extend %fa;
  }
}
//new: <a class="my-button">this is a button</a>
//old: <a><i class="fa fa-star"></i> this is a button</a>
```

### DIFF
--- a/scss/_icons.scss
+++ b/scss/_icons.scss
@@ -6,15 +6,15 @@
 @each $var in $fa-vars {
   //create icon placeholders
   %fa-#{nth($var, 1)} {
-    &:before {
-      content: nth($var, 2);
-    }
+    content: nth($var, 2);
   }
   
   //create icon classes if enabled
   @if $fa-show-classes == true {
     .#{$fa-css-prefix}-#{nth($var, 1)} {
-      @extend %fa-#{nth($var, 1)};
+      &:before {
+        @extend %fa-#{nth($var, 1)};
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


intensr:
Is it possible to add just one new file (e.g. `_icons-silent` or so) with all fa classes as silent classes for scss? This approach is very nice to avoid bloated code if used as extend wisely e.g.

`.#{$fa-css-prefix}-glass:before { content: $fa-var-glass; }`

as

`%#{$fa-css-prefix}-glass:before { content: $fa-var-glass; }`

or simply

`#{$fa-css-prefix}-glass:before { content: $fa-var-glass; }`


At the moment there is no possibility to even hack this.
Regards, Dave

PR #3268, opened 2014-04-17
Corner support for stacked icons

jakofranko:
I built out a little CSS to support the fa-stack-corner-* class for stacked icons.

You can specify an icon to be in the nw, ne, sw or se corner. The corner icons will size dynamically if placed next to a fa-stack-2x icon, or if given the class fa-inverse or if placed next to another icon with the classname fa-inverse.

I also added support for black and white text-shadows based on whether the corner icon or it's sibling is normal color or inversed.


zuck:
+1

codynguyen1116:
+1

jakofranko:
This is 2 years old...gonna close out.

Issue #671, opened 2013-01-13
Rendering issue: icon-spin fa-spin shake/wobble (chrome firefox)


Katuva:
When using data-loading-text, for example:

data-loading-text="<i class='icon-refresh icon-spin'></i> Logging in..."

And then when the button is clicked and the "loading" property is set, the icon will wobble and shake as it spins.

Although, if you include the spinning icon on the button as normal, for example:

<button type="button" id="login_btn" class="btn btn-primary"><i class='icon-refresh icon-spin'></i> Log in</button>

It spins fine.

I've only encountered this bug on Firefox (currently version 18), works perfectly on Chrome.

**edit by tagliala**
Firefox related issue: #3451

skoub:
The problem seems to come from the icon itself. If you take the Icomoon spinner version, it does'nt wobble at all. It works well on Chrome 39, Windows and IE11, Windows. I don't know for the other browsers.

http://jsfiddle.net/cxxzh0uf/

tagliala:
@skoub 

The approach of icomoon of keeping icons all of the same size seems better for this purpose but it isn't enough.

http://jsfiddle.net/tagliala/cxxzh0uf/2/

![image](https://cloud.githubusercontent.com/assets/556268/5607297/091e0aa8-9456-11e4-966d-e356f6ac1cc4.png)

If I inspect the `:before` pseudoelement, I notice that fontawesome is "weird"

![image](https://cloud.githubusercontent.com/assets/556268/5607303/35639d30-9456-11e4-8ac5-93180d3cdce6.png)

![image](https://cloud.githubusercontent.com/assets/556268/5607305/46e3c864-9456-11e4-874e-7197284305b9.png)

I think that this is related to the font itself and there is nothing we can do via CSS.

Dave should take a look at the font and inspect fonts generated by Icomoon to try to understand what is the best way to fix this issue.

We can't accept pull requests with edits to the .otf file, but we will appreciate some help on this matter. If someone here is able to take a look at the font and identify the problem is very welcome

Sorry for my English, it is also late here :)

luxzg:
@tagliala thanks for linking fa-clock-o issue here as well, it's the same issue as rest are mentioning.
[...]

PR #4130, opened 2014-08-13
New spinning effect for fa-spinner

Seb33300:
This is a new effect more appropriated for the fa-spinner icon.
I don't know how to compile scss and less into css in the main directory.
Nothing specified about that in the guidelines.


skoub:
The problem seems to come from the icon itself. If you take the Icomoon spinner version, it does'nt wobble at all.
http://jsfiddle.net/cxxzh0uf/

tagliala:
@skoub please post your comment about the spin wobble here: #671, I'm interested and I have a clue

skoub:
@tagliala, i added the comment to the #671 thread.

Seb33300:
Better implementation in #4456

Issue #2858, opened 2014-02-08
Feature requests: fa-rotate-45


magoo-pl:
Yesterday I need to do a close button and I want to use font-awesome to do it. And there was no cross icon or something like that. I simply used plus icon and rotate it 45deg with additional class.

I want to suggest a "fa-rotate-45" class for icons.

sarciszewski:
> IE8/9 really sucks on 45deg rotations:

> IE8/9 really sucks 

FTFY -- is IE 8/9 support going to be revoked in a future version of Font Awesome?

shunsukeaida:
Not sure FA's support policy, but I'm concern it's akin to ask when Windows 7 will die...

KerriLea:
+1 for fa-rotate-45

zuhloobie:
I use this for 45 rotations - just add onto font-awesome.min.css file:

`
.fa-rotate-45 {-webkit-transform:rotate(45deg); -moz-transform:rotate(45deg); -ms-transform:rotate(45deg); -o-transform:rotate(45deg); transform:rotate(45deg);}
`

Then add `fa-rotate-45` to inline span code like `<span class="fa fa-hand-o-right fa-rotate-45"></span>`

javiros:
@zuhloobie thank you!

isuvorov:
+1

coeus01:
**+1**

My1:
@shunsukeaida nah I rather think until vista dies. Win7 got IE Updates and older IE versions than the latest available on a given system are dropped out of MS's support corner.
and the only OS with IE9 as latest is Vista.

tagliala:
Closing here in favour of #8515

PR #4210, opened 2014-08-23
Add rotate for 45deg, 135deg, 225deg, 315 deg

hebbet:
Add rotate for 45deg, 135deg, 225deg, 315 deg

See also PR #3771 and #2858


rmullaney77:
See https://github.com/rmullaney77/Font-Awesome/commit/21054c55bb794502529bd52aec0f51a01deebe56

For IE6/7 support in addition to current commit for IE8, add "filter:" in addition to "-ms-filter:" properties ;)

PR #4458, opened 2014-09-17
Firefox shows text selection cursor when hovering over fa-icon

ramonakira:
Firefox shows text selection cursor when hovering over fa-icon. I added cursor: default; to just show the default cursor of the OS, see here: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

### DIFF
--- a/less/core.less
+++ b/less/core.less
@@ -8,4 +8,5 @@
   text-rendering: auto; // optimizelegibility throws things off #1094
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  cursor: default;
 }


fredgate:
I agree with @ramonakira : font-awesome are icons, certainly stored in a font, but there are icons. The user see icons, not a font, so this should be rendered as an icon.
The jsfiddle of @ramonakira is good : set a default cursor for .fa, and set a pointer cursor when the .fa is inside a link.

tagliala:
Still against.

Images inside links show `cursor: pointer` when hovered: http://jsfiddle.net/cb0s4wjb/67/

If you have special needs, please edit your css

fredgate:
It does not seem a special need, but a common behavior. I never saw a image with a cursor as a text selection. But this is what occurs with font awesome icon on Firefox.
This can be resolved with a CSS rule; and when icon is inside a link, as showed by @ramonakira, a simple second rule is used.

```
.fa {
    cursor: default;
}
a .fa:hover {
    cursor: pointer;
}
```
Icons always display as it on all sites.
I can add it to my css to customize font awesome, but I think that it could (should) be present base in font awesome.

PR #5209, opened 2014-12-13
Fix Jekyll build command

williamboman:
Now uses local `lessc` npm binary when building with Jekyll. Also added `less` to package.json.

The `--yui-compress` option was deprecated in `less#1.5`, so now uses `--compress` (is/was there a difference between these two?).

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,9 +40,11 @@
       "url": "http://opensource.org/licenses/mit-license.html"
     }
   ],
-  "dependencies": {
+  "dependencies": {},
+  "engines": {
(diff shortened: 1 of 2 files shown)


tagliala:
@williamboman this was closed because I removed the 4.3.0-wip branch but this less thing is hurting me.  I really want this fixed but I need to ask Dave before

williamboman:
Yeah it took me quite a while to figure out what was wrong when I initially cloned the repo and tried to build with jekyll.. This was the best I could come up with quickly.

PR #5277, opened 2014-12-23
Feature: jumping animation for .fa-spinner.fa-spin

spieker:
This way it looks more like the way you expect .fa-spinner to be animated. What do you think?


tagliala:
@spieker is this something like the new "pulse" animation of FA 4.3.0?

spieker:
@tagliala Yes, it is

tagliala:
thanks for the PR :+1:

PR #5379, opened 2015-01-12
Added component to composer.json

gintsmurans:
Components are really useful when composer vendor directory is not directly accessible by webserver, this way I can specify exact directory where those files should be put. More about it here: https://github.com/RobLoach/component-installer


tagliala:
@gintsmurans since we probably need a 4.3.1 version, could you please push against 4.3.1-wip branch?

robmadole:

– FA team

PR #5460, opened 2015-01-23
Wrap the core styling into a mixin for reusability

msikma:
It seems better to me to have a mixin for the core styles so that it becomes possible to easily turn other elements into icons (notably pseudo-elements), without having to locally duplicate the styles. This minor change makes that possible by letting people use `@include font-awesome-icon();`.

There is no change to the compiled CSS code. This code works on SassC and Ruby Sass.

### DIFF
--- a/scss/_core.scss
+++ b/scss/_core.scss
@@ -1,11 +1,17 @@
 // Base Class Definition
 // -------------------------
 
-.#{$fa-css-prefix} {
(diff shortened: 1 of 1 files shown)


msikma:
Sorry, I see now that this is the wrong branch. This is a modification of the master branch. I'll remake the pull request, properly this time. My apologies.

Issue #5198, opened 2014-12-12
Provide a mixin for each icon


unwiredbrain:
It would be extremely useful to have mixins for each icon, something like the following:

``` scss
.step {
  @include fa-icon;
  @include fa-icon-circle-o;
}

.step--is-current { @include fa-icon-circle; }
.step--is-complete { @include fa-icon-check-circle; }
```

A similar behaviour can be obtained using `@extend` but it has just too many drawbacks, being so powerful that sometimes it can be dangerous (mangled source order & co.) as many articles and blog posts don't fail to outline.

Here's @csswizardry take on the topic: [When to use @⁠extend; when to use a mixin ](http://csswizardry.com/2014/11/when-to-use-extend-when-to-use-a-mixin/)

mindplay-dk:
In LESS, any class works as a mix-in, so `icons.less` could be modified to work as mixins quite easily - in fact, it's just a syntax tweak away, so this:

    @{fa-css-prefix}-check:before { content: @fa-var-check; }

Becomes that:

    @{fa-css-prefix}-check { &:before { content: @fa-var-check; } }

This would enable you to create more semantic CSS rules, e.g. using `import (reference)` for the font-awesome files, so that you get only the properties and CSS rules that are actually in use in your stylesheet, rather than embedding the whole library.

sensibleworld:
Doing a little management in Github and (batch) closing out potentially stale issues. If the above still needs attention, please let us know and we’ll re-open!

PR #5468, opened 2015-01-24
Wrap the core styling into a mixin for reusability

msikma:
This modification creates a mixin for the core styles so that it becomes possible to easily turn other elements into icons without having to locally duplicate the styles. This is particularly useful when working with pseudo-elements. People would be able to simply use `@include font-awesome-icon();` in any element.

I compiled the modified version and the original version, and there is no difference in the output.

### DIFF
--- a/less/_core.less
+++ b/less/_core.less
@@ -2,10 +2,5 @@
 // -------------------------
 
 .@{fa-css-prefix} {
-  display: inline-block;
-  font: normal normal normal 14px/1 FontAwesome; // shortening font declaration
-  font-size: inherit; // can't have font-size inherit on line above, so need to override
-  text-rendering: auto; // optimizelegibility throws things off #1094
-  -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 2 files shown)


tagliala:
Thanks for this one.

The PR at the moment is wrong and cannot be merged.

You should edit both less and scss files within the src folder, as per contributing guidelines: https://github.com/FortAwesome/Font-Awesome/blob/master/CONTRIBUTING.md#pull-requests

You should also put the mixin in the proper files: https://github.com/FortAwesome/Font-Awesome/tree/master/src/assets/font-awesome

This mixin would probably be useful for other people.

Refers to #5198

msikma:
Thanks for the feedback, I will modify the request soon.

This would solve a problem that I had writing my own framework, where I wanted to have a specific FA icon in a search bar (among other things), without people having the need to add the class.

msikma:
Closing this now since it doesn't seem necessary anymore with a core styles mixin already being available in the `src/` directory on master.

PR #5655, opened 2015-02-12
Add woff2 to component.json

tagCincy:
Recently added woff2 file was not added to component.json.  Was causing issues when using services like rails-assets.org.


tagliala:
@tagCincy thanks.

Is there something else we can do?

tagliala:
FFFFFF

tagliala:
@tagCincy please push against the proper branch: `4.3.1-wip`

tagCincy:
PR #5657

PR #5679, opened 2015-02-15
Add a 'hot air balloon' icon

loanlaux:
I think it would be a good idea to implement a 'hot air balloon' icon, styled the Font Awesome way. We could go for something like this one (perhaps with less curves on the higher part, without any human in the pod and without any loose cable between it and the balloon itself):

Vectorial balloon

Just dropping this idea by here. I just hope people will feel the need for this icon, too.


tagliala:
Hi, thanks for this PR but I can't merge this one:

https://github.com/FortAwesome/Font-Awesome/blob/master/CONTRIBUTING.md#pull-requests

> At the moment we are not accepting pull requests containing icons

At the moment there is no request about a hot air balloon, if someone ask for this I will point they to your fork, thanks again!

PR #5708, opened 2015-02-18
Added tick and cross aliases for check and times

tjbp:
I occasionally find myself using the check and times icons for feature lists or other tables of boolean values. In British English a check is more commonly called a "tick", and its opposite is a "cross". I often can't remember to use "times", and I don't even know the American English name for a cross (an "x"?).

I've not recompiled the fonts and minimised the CSS, since that'll quickly cause a merge conflict, but I can take care of it if needed.


tjbp:
@davegandy so that's a no then?

PR #5750, opened 2015-02-22
.fa { text-indent: initial; }

Mr21:
Hi :)

I use *Font-Awesome* every day, and every time I have to add this in my code:

```css
.fa {
    text-indent: initial;
}
```

The probleme is the `text-indent` property is set to `inherit` by default.
So if I choose to put some `<i class="fa fa-*"></i>` in a `<p style="text-indent: 1em;">`.
All the `<i>` inside will have a margin before the icon and it's not good ^^

I have put an example here: http://mr21.fr/fa/text-indent/


tagliala:
Interesting, thanks.

This could break things, so we should include it in 5.0.0

Please push against `5.0.0-wip` branch. Please also make sure to edit the proper files that are under the `/src` folder, as per our contributing guidelines: https://github.com/FortAwesome/Font-Awesome/blob/master/CONTRIBUTING.md#pull-requests

Can't merge this one so I'm closing here

PR #4234, opened 2014-08-27
Added version indicators for new icons to cheatsheet

rmm5t:
In response to #4230.

If an icon is newer than the current major version, show a version badge
to indicate which version of FontAwesome this icon was added.

Closes #4230

font_awesome_cheatsheet

PR #5787, opened 2015-02-25
Added version indicators for new icons to cheatsheet

### LINKED ISSUE #4230: Better indication of versions within the cheatsheet
Repost from issue #4191:

@tagliala Right, my bad, I was using 4.1.0, updating to 4.2.0 solved the problem. Thanks.

(Icons remained Minion Pro squares even after specifying them as FontAwesome...irrelevant now.)

As for feedback, it would be nice to indicate "new" icons in the cheatsheet. As a designer I use 99% of the time only the cheatsheet as a direct source for Photoshop. I don't check other pages of the site and I would guess other designers do not either. It didn't struck me that I need to update to a new version of the font because I didn't notice there is a new version. From a UX point of view and my perspective of using the site, having two separate pages Icons and Cheatsheet seems a bit redundant, imho.

rmm5t:
If an icon is newer than the current major version, show a version badge
to indicate which version of FontAwesome this icon was added.

See #4234
Closes #4230

/cc @tagliala

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -21,6 +21,7 @@ icon_destination: icon         # Relative to destination
 fontawesome:
   version:        4.3.0
   minor_version:  4.3
+  major_version:  4
   doc_blob:       v4.3.0
   url:            http://fontawesome.io
   legacy_url:     http://fortawesome.github.com/Font-Awesome/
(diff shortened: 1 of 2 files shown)


tagliala:
:+1: perfect, thanks

PR #5901, opened 2015-03-09
YML fix

ajbrun:
Taken from 4.3.1 wip to resolve a missing aliases header in the yml.


tagliala:
@ajbrun thanks

This is already fixed in the 4.3.1-wip branch.

Please be patient, we recently changed our contributing guidelines and the fix was already pushed against that branch. I will not merge this one but it should not happen again in the future

PR #5934, opened 2015-03-11
Fix link to spinning examples href to be #animated instead of #spinning

timlevett:
`#spinning` doesn't exist currently in the examples page, however `#animated` does.


tagliala:
thanks

PR #6020, opened 2015-03-20
Sass: List var of all icons

CupOfTea696:
Adds a list variable of all icons to easily loop through


ketan:
+1 - it would be great to see this change, so that font awesome is more usable with sass mixin frameworks.

I'm happy to rebase and submit a new PR if the maintainers are OK with this change, OR have a better idea to refactor the variables file so it's easier to "lookup" icons.

XiBOR:
+1

drmonkeyninja:
This seems like a really good idea as it would surely make the Sass not only more usable but also easier to maintain as adding new icons would be as simple as adding a new icon to the list. Is there any chance of this being revisited and actually accepted into the master branch?

Matthew-Bonner:
Can you change the title from SASS to Sass. See:
https://github.com/FortAwesome/Font-Awesome/pull/8170

robmadole:

– FA team

PR #6174, opened 2015-04-07
Added filter for 'ban' / Sanitized format

Prismapanda:


tagliala:
Hi. Thanks but this is already fixed in 4.3.1-wip branch. Sorry but we changed our contributing guidelines recently and this is why the fix is not included in the master

closing here

Issue #5082, opened 2014-11-27
Apply fa-spin to :before pseudoelement


chderen:
Hi,

should the spin apply only on the icon it's self?
i mean use .fa-spin:before instead of .fa-spin.

Thanks

Issue #2695, opened 2014-01-04
Suggestion: .fa-before


Traik:
In font-awesome.css there is the following class:

<pre>.fa {
  display: inline-block;
  font-family: FontAwesome;
  font-style: normal;
  font-weight: normal;
  line-height: 1;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}</pre>

and if someone wants to add an icon to a link for example: `<a href="" class="fa fa-some-icon">link</a>` then the word `link` has `font-family: FontAwesome;` and there is no way to change that other than by adding another class: `.fa-before`.

Therefore, I suggest adding `.fa-before` to the main package. It would apply `font-family: FontAwesome;` only to the `:before` pseudo-element:

<pre>.fa-before:before {
  display: inline-block;
  font-family: FontAwesome;
  font-style: normal;
  font-weight: normal;
  line-height: 1;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}</pre>

New usage: `<a href="" class="fa-before fa-some-icon">link</a>`

Mr0grog:
I’m running into the same issue—for me, it’s dramatically better to not have those `<i>` tags all over. I’m working around it in my CSS by overriding the `.fa` class to do what it does on `fa:before` instead. I’m not sure if, general-purpose-wise, that’s a workable solution, but it seems like the best way to me. It would be great to have FontAwesome work that way by default.

sensibleworld:
Doing a little management in Github and (batch) closing out potentially stale issues. If the above still needs attention, please let us know and we’ll re-open!

PR #6223, opened 2015-04-14
Apply all the specific CSS on the pseudo-element (:before)

Mr21:
Hi :)

In this page I use the true FontAwesome **4.3.0**.
[http://mr21.github.io/_/fa/test-pseudo-elements/**before**.html](http://mr21.github.io/_/fa/test-pseudo-elements/before.html)
There are two columns, on the left I've pasted all your examples I found at:
http://fontawesome.io/examples/

And on the right I change each example.
For example, I changed this :
`<i class="fa fa-camera-retro"></i> fa-camera-retro<br/>` by this:
`<i class="fa fa-camera-retro"> fa-camera-retro</i><br/>`

And this :
`<li><i class="fa-li fa fa-check-square"></i>List icons</li>` by this:
`<li class="fa-li fa fa-check-square">List icons</li>`

We can see, none works really good on the right.
At the same time **this is not how we should use FontAwesome**, I understood.

But why not allow the two notations.
Many times we have to create another DOM element for... nothing.
Actually, something like this:
`Click here: <a href="#" class="fa fa-link"></a>` will works and it's prettier than:
`Click here: <a href="#"><i class="fa fa-link"></i></a>`

But this rules not works for the all case.

**My change doesn't break anything**, it just allows another notation for saving a few DOM elements sometimes. You can check here:
[http://mr21.github.io/_/fa/test-pseudo-elements/**after**.html](http://mr21.github.io/_/fa/test-pseudo-elements/after.html).

I add a `:before` behind every CSS rules **except** after: `.fa-ul { ... }` and `.fa-ul > li { ... }`

PS: I firstly pushed only the `css/font-awesome.css`.

### DIFF
--- a/css/font-awesome.css
+++ b/css/font-awesome.css
@@ -11,7 +11,7 @@
   font-weight: normal;
   font-style: normal;
 }
-.fa {
+.fa:before {
   display: inline-block;
   font: normal normal normal 14px/1 FontAwesome;
   font-size: inherit;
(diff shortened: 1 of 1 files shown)


Mr21:
@davegandy, @tagliala what do you think about this?

tagliala:
> At the same time this is not how we should use FontAwesome, I understood.

IMHO If you have special needs, you should customize your own build

There should be issues related  to before pseudoelements: https://github.com/FortAwesome/Font-Awesome/issues/5082#issuecomment-66043325

Refers to #2695 #1729

Mr21:
I don't have a special needs. It's just an idea to simplify the usage of Font-Awesome.
This change not breaking anything, It allows to write less HTML for doing the samething.

Mr21:
Closed after 6 months without any answer...
What is the deal @davegandy ?

Tell us if you absolutly refuse any contribution...

leoyli:
I also don't get it why FA have to be limited in `<i>` tag...

Issue #6113, opened 2015-03-30
Configurable @font-face { font-family: $fa-font-family }


chrisdpeters:
I am building an HTML interface that must be mixed in with customers' HTML interfaces. Customers could end up including their own version of Font Awesome, thus clashing with my own. I'd like to be able to namespace the `font-family` within the `@font-face` declaration so that the version of Font Awesome I'm using doesn't clash with my customers'.

I'd be willing to figure out how to do this and submit a pull request, and soon. Do you find this to be a sensible variable to add to the Less and Sass versions? If so, which branch would you prefer that I push this to?

tagliala:
Hi, 

Sorry, but this is a wontfix at the moment.

If you need to change the name of the font, you can use your own `_core.scss`/`_core.less`

Closing here

PR #6386, opened 2015-05-05
add(core) font name variable

Grawl:
~ move font name into variable to allow of using customized FontAwesome
with different font name

### DIFF
--- a/less/core.less
+++ b/less/core.less
@@ -3,7 +3,7 @@
 
 .@{fa-css-prefix} {
   display: inline-block;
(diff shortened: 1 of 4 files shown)


tagliala:
Fix #6113 but this PR is not editing the source files

Also, I'm going to label #6113 as wontfix at the moment, sorry!

Grawl:
sorry :cat: 

but why `wontfix`?

tagliala:
@Grawl sorry, I will let you know the reason. Dave is probably having a really busy time

PR #6618, opened 2015-06-01
Create LICENSE

bhamodi:
I propose that we use the MIT license based on the reasons provided here: http://choosealicense.com/


williamboman:
https://github.com/FortAwesome/Font-Awesome#license

bhamodi:
Oh good catch!

Issue #6552, opened 2015-05-25
Loading font-awesome with webpack, wrong content attribute


stevewillard:
Anyone know why when I load font-awesome into my site via webpack, all the content attributes of my icons are wrong? Maybe it's an encoding issue? Here's my config:

```
require("font-awesome/scss/font-awesome.scss");
```

```
module: {
        loaders: [
            { test: /\.jsx?$/, exclude: /node_modules|bower_components/, loader: "6to5-loader"},
            { test: /\.scss$/, loader: "style!css!sass?includePaths[]=" + bourbon },
            { test: /\.woff(2)?(\?v=[0-9]\.[0-9]\.[0-9])?$/, loader: "url-loader?limit=10000&minetype=application/font-woff" },
            { 
                test: /\.jpe?g$|\.gif$|\.png$/,
                loader: "url-loader" 
            },
            { test: /\.(ttf|eot|svg)(\?v=[0-9]\.[0-9]\.[0-9])?$/, loader: "file-loader" }
        ]
    }
```

If you inspect the DOM, the stylesheet gets added but but the content attributes look like this:
![screen shot 2015-05-25 at 1 41 43 pm](https://cloud.githubusercontent.com/assets/11420/7800474/dcb2d5fa-02e3-11e5-9bca-9fddd5be57d0.png)

But, they look correct in my node_modules folder. Any ideas?

johnbendi:
Did I just see minetype?‎ Instead of mimetype.                                                                                                                                                                                                                                                                                                                                         Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: Steve WillardSent: Monday, May 25, 2015 6:42 PMTo: FortAwesome/Font-AwesomeReply To: FortAwesome/Font-AwesomeSubject: [Font-Awesome] Loading font-awesome with webpack, wrong content attribute (#6552)Anyone know why when I load font-awesome into my site via webpack, all the content attributes of my icons are wrong? Maybe it's an encoding issue? Here's my config:

require("font-awesome/scss/font-awesome.scss");


module: {
        loaders: [
            { test: /\.jsx?$/, exclude: /node_modules|bower_components/, loader: "6to5-loader"},
            { test: /\.scss$/, loader: "style!css!sass?includePaths[]=" + bourbon },
            { test: /\.woff(2)?(\?v=[0-9]\.[0-9]\.[0-9])?$/, loader: "url-loader?limit=10000&minetype=application/font-woff" },
[...]

Issue #6489, opened 2015-05-16
W3 Jigsaw fails to validate fontawesome.min.css


bitzl:
fontawesome.min.css fails validation in several points:

http://jigsaw.w3.org/css-validator/validator?uri=http://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css&warning=0&profile=css3

While some issues address special browser-specific stuff like "@-webkit-keyframes", there are also issues regarding missing semicolons and parse errors.

tagliala:
Sorry, we need to support some browsers that needs custom prefixes and non-standard properties.

If you are not interested in those browser, please consider a custom build of fontawesome

Hope it helps, closing here

bitzl:
Wow, that was fast. But I did *not* complain on custom prefixes and non-standard properties. These are pretty ok if they help to support more browsers.

My point are the "missing semicolons and parse errors" as reported by Jigsaw. If you are interested in this issue, I'd be willing to help.

tagliala:
Unfortunately they are a consequence of the `filter:` property

Try to validate via direct input:
```css
.fa-rotate-90 {
  filter: progid:DXImageTransform.Microsoft.BasicImage(rotation=1);
  -webkit-transform: rotate(90deg);
  -ms-transform: rotate(90deg);
  transform: rotate(90deg);
}
```

and then:
```css
.fa-rotate-90 {
  -webkit-transform: rotate(90deg);
  -ms-transform: rotate(90deg);
  transform: rotate(90deg);
}
```

PR #6728, opened 2015-06-11
fixes issue #6552

bsara:
See issue #6552 comments for details on why the issue was occurring and what changes needed to happen to fix the issue.


tagliala:
Thanks

I can't merge this on the master branch or in the 4.3.1-wip branch because it breaks backward compatibility.

My thouhgts on this one:

I would go for `$var-invalid: unquote('"\\f000"');`.

Yes, this is ugly and it will increase the file size of _variables.scss but we will not going to break backward compatibility and we can still use `content: $fa-icon-whatever`, without the `prepend-slash`

bsara:
in what way does this break backward compatibility?

tagliala:
it will break backward compatibility for users that are using something like

http://jsfiddle.net/tagliala/w595xhLo/

```scss
a {
   &:before {
     font-family: FontAwesome;
     content: $fa-var-link;
   }
   &[target="_blank"]:before {
     content: $fa-var-external-link;
   }
}
```

```html
<a href="#">Link</a><br>
<a href="#" target="_blank">External Link</a>
```

bsara:
But that code won't even work upon compilation. I see that it works in that particular JS Fiddle, but in practice it does not work that way. I don't know which version of SASS is being used by JS Fiddle either, I want to say that I remember that this used to not be an issue in SASS but is now. If you create a file called `test.scss` at the root of the SCSS files, then add the code below and compile via the terminal, you'll see that this issue cannot be avoided and that though these variables can't be used directly, they aren't even working currently, so what's the point of keeping backwards compatibility if the currently solution doesn't even work?

**test.scss**

```scss
@import 'variables';

.my-class:before {
  content: $fa-var-link;
}
```

It would seem prudent to provide a mechanism for getting a glyph character. Perhaps a function called `fa-get-glyph($glyph-code)`? Then, to get the glyphs that you reference above, you would have the following: `content: fa-get-glyph($fa-var-external-link);` This would provide a way to still access those glyphs directly but will allow for someone to know that they are getting what they expect to get, rather then just a hex value. However, anyone who has any amount of good experience with SASS/SCSS wouldn't use these unless they needed the glyph for a property who's selector does not contain `:before`. They would/should, instead extend a placeholder or the class itself so that you are guaranteed that you always have what you think that you're going to get. (I.E. `@extend .fa-external-link;`)

tagliala:
There is definitely an issue here, but here it works even if the css is not using the proper escaping.

My starter app suffers of the same issue: http://ruby2-rails4-bootstrap-heroku.herokuapp.com/ but it works. Maybe some encoding stuff involved?

> It would seem prudent to provide a mechanism for getting a glyph character. Perhaps a function called fa-get-glyph($glyph-code)? 

That's ok but I don't want to intentionally break backward compatibility. I'm going to assign this to @davegandy

tagliala:
@bsara 

I would go for `$fa-var-icon: unquote('"\\f000"');` for 4.x.x and change this in 5.0.0. what do you think?

bsara:
@tagliala

That would make sense to add something like this in a major release. However, for the time being, the `unquote` really won't make any difference. So, perhaps this pull request should be added to  the 5.0.0-wip branch instead?

bsara:
@tagliala I just realized something today! Would it be against the code standards of the project to just remove the quotes from the icon codes? (IE `\f000` instead of `"\f000"`)

I realized that this will work just fine but will retain backwards compatibility. Essentially, that's what the `unquote` function is turning the string into anyway!

tagliala:
@bsara thanks for the heads up!

Unfortunately I didn't get the point... Could you please edit your old sassmeister gist here http://sassmeister.com/gist/3bfad69a4ca6e34909cf to show what we can do?

bsara:
@tagliala done. Also, I've already made the change in my local copy of the project and have successfully built as well, I'm ready to commit and submit a pull request assuming that you will accept the change.

tagliala:
Unfortunately this produces an invalid css file

tagliala:
I just noticed that the jsfiddle sass compiler turns 

```scss
// ----
// Sass (v3.4.14)
// Compass (v1.0.3)
// ----

@function prepend-slash($value) {
  @return unquote('"\\#{$value}"');
}

// calling `prepend-slash` is only necessary when using a variable

$var-valid0: "f000";
$var-valid1: unquote('"\\f000"');
$var-invalid: "\f000";

.valid0:before {
  content: prepend-slash($var-valid0);
}

.valid1:before {
  content: $var-valid1;
}

.invalid:before {
  content: $var-invalid;
}

.no-var:before {
  content: "\f000";
}
```

into

```css

    .valid0:before {
  content: "\\f000"; }

  .valid1:before {
    content: "\\f000"; }

    .invalid:before {
      content: "\f000"; }

      .no-var:before {
        content: "\f000"; }

```

So it will break stuff for someone :cry:

tagliala:
After reading this comment, https://github.com/sass/sass/issues/659#issuecomment-111446446

I think we should drop those variables in 5.0.0. 

> They would/should, instead extend a placeholder or the class itself so that you are guaranteed that you always have what you think that you're going to get. (I.E. `@extend .fa-external-link;`)

bsara:
@tagliala I actually agree with you, in fact, I've already written the code for implementing placeholders in the project, I've been using that code for ever a week and it's been working wonderfully. However, I for the sake of better merging, I wanted to submit this fix first, then submit my placeholder changes.

Though, there is great advantage to have these values stored in variables as well I think. I personally feel that both should be available. And even if the variables are done away with in 5.0.0. The bug will still exist in the next/current version and, in my opinion, it's quite the bug. I wonder if, at the very least, we fix the bug for 4.3.1 and continue the discussion of placeholders for 5.0.0? For the time being I will submit a pull request of the bug fix so that you can at least see how small the change is to fix the bug. But I won't be able to submit it until about 45-60 minutes from now.

tagliala:
My fear is that the fix will break old sass libraries :disappointed: 

I don't know what is the worst choice

1. Invalid CSS via `$var: \f000` (we have many reports for invalid css, like #6489, for just a couple of rules)
2. `unescape`, that will break some (old?) sass libraries
3. new mixin that will break backward compatibility *and* some sass libraries

I would to hear something from Dave, but he should be extremely busy :confused:

tagliala:
@bsara another idea

1. Create a new file, let's say `_unquoted_variables.scss`, like

  ```scss
  $fa-var-icon: unquote('"\\f000"');
  ```

2. Add `!default` to `$fa-var-icon` in `variables.scss`
  ```scss
  $fa-var-icon: "\f000" !default;
  ```

3. Document the issue, asking people affected by this issue to require `_unquoted_variables.scss` until the release of 5.0.0

In this way we could fix the issue and keep backward compatibility

bsara:
@tagliala that would work. Also, in 5.0.0 it would probably be a good idea to pick a minimum Sass version to support. It seems that this workaround is only needed in Sass [v3.4.0+](http://sass-lang.com/documentation/file.SASS_CHANGELOG.html#340_18_august_2014). (You can try changing the version that SassMeister is compiling my example at and you'll see a difference in output.

The only issue I can think of is the fact that people will be able to accidentally change the variable values...but I doubt that's an issue since you can essentially do that now. You just need to set the variable after the font awesome SCSS files are included rather than before or after.

I'll hack out that change right now and let you know how it's looking. Shouldn't take long.

Just to clarify: Do you agree that for 5.0.0 the `_unquoted-variables.scss` can go away and just using `unquote('"\f000"')` when setting the variable values?

tagliala:
Thanks!

> Just to clarify: Do you agree that for 5.0.0 the _unquoted-variables.scss can go away and just using unquote('"\\f000"') when setting the variable values?

Yes, or we can use the mixin you suggested. We can do the best we can for 5.0.0 without thinking about backward compatibility :)

bsara:
Awesome! I'll submit the request now.

tagliala:
Improved gist: https://www.sassmeister.com/gist/984c3d8842170a8a21c1a787e4a25a86

ssp5zone:
> @bsara another idea
> 
> 1. Create a new file, let's say `_unquoted_variables.scss`, like
>    ```css-scss
>    $fa-var-icon: unquote('"\\f000"');
>    ```
> 2. Add `!default` to `$fa-var-icon` in `variables.scss`
>    ```css-scss
>    $fa-var-icon: "\f000" !default;
>    ```
> 3. Document the issue, asking people affected by this issue to require `_unquoted_variables.scss` before `font-awesome.scss` until the release of 5.0.0
> 
> In this way we could fix the issue and keep backward compatibility

This would have really helped. Nonetheless, I could not upgrade to 5/6 alpha because my current app is quite big and old, and Apache Tomcat's serializer just messes up things. So, I ended up overriding the variable.scss as follows,

```scss
$fa-font-path:"~font-awesome/fonts";

@import "overrides/font-awesome/variables";
@import "~font-awesome/scss/mixins";
@import "~font-awesome/scss/path";
@import "~font-awesome/scss/core";
@import "~font-awesome/scss/larger";
@import "~font-awesome/scss/fixed-width";
@import "~font-awesome/scss/list";
@import "~font-awesome/scss/bordered-pulled";
@import "~font-awesome/scss/animated";
@import "~font-awesome/scss/rotated-flipped";
@import "~font-awesome/scss/stacked";
@import "~font-awesome/scss/icons";
@import "~font-awesome/scss/screen-reader";
```

PR #3386, opened 2014-05-09
Add fa-xl size class, between fa-lg and fa-2x

johnnyshields:
Add fa-xl size class, between fa-lg and fa-2x

PR #6839, opened 2015-06-29
SASS/LessCSS refactor

zakmac:
* This work comes from enhancements made to support [erikflowers/weather-icons#82](https://github.com/erikflowers/weather-icons/pull/82) that can be applied to the FA project.
* Includes a similar enhancement as requested in #6020.
* Includes a similar enhancement as requested in #3386.

These changes allow more flexibility with icons and sizing/rotation configuration, without introducing deprecations.

## Changes specific to `./scss`

**font-awesome.scss**
- added `functions.scss`

**functions.scss**
- added convenience functions
  - `fudge` fudge a number from 0-6 decimal places
  - `unitless` simple hack to parse a number from a unit

## Changes applicable to both `./less` and `./scss`

**version**
- bump to 4.4.0

**variables.less**
- added variables:
  - deprecated explicit icon variables in favor of an array value (`@fa-var-glyphs`)
  - `@fa-rotation-increment`
  - `@fa-var-scales`
  - `@fa-var-multipliers`

**icons.less**
- removed explicit definitions in favor of a loop iterating `@fa-var-glyphs`

**mixins.less**
- removed requirement to pass `@rotation` to `.fa-icon-rotate()`, it is now calculated
  - calculation required since `@fa-rotation-increment` is configurable, meaning not every rotation may adhere to 90 degree increments requiring `@rotation`

**rotated-flipped.less**
- rotations are now calculated by a loop iterating `@fa-rotation-increment`
- the ie fix is now calculated using a loop iterating only the rotation selectors requiring intervention
  - this is required now that `@fa-rotation-increment` exists

**larger.less->sizes.less**
- file has been renamed to reflect intent
- sizes are now calculated by a loop iterating `@fa-var-scales`
  - this new method supports the existing `.fa-lg` (+33%) selector, as well as introducing: `.fa-xsm` (-66%), `.fa-sm` (-33%x), `.fa-default` (100%), `.fa-xlg` (+66%)
- multipliers are now calculated by a loop iterating `@fa-var-multipliers`
  - this new method supports all existing multipliers (`.fa-2x`, `.fa-3x`, `.fa-4x`, `.fa-5x`)

**misc.**
- minor code cleanup


zakmac:
There is an issue in `rotate-flip.[scss/less]` where the MSIE fixes are all declared as separate variables. The only solution I've got is to have each option extend a base selector with that property, however then you are left with a junk property, example below:

```css
.ie-rotation-fix {
  filter: none;
}
```

**SASS extend**
```scss
@for $i from 0 through 359 {
  @if $i == 0 or $i % $fa-rotation-increment == 0 {
    $rotation-selector: #{$fa-css-prefix}-rotate-#{$i};
    // check if this item needs some ie love
    @if $i == 0 or $i % 90 == 0 {
      .#{$rotation-selector} {
        @extend .ie-rotation-fix;
      }
    }
  }
}
```

**LessCSS extend**
```less
:root  {
  .calculateIEFixes(@i) when (@i >= 0) {
    // iterate
    .calculateIEFixes((@i - 1));
      @z: ~"@{wi-css-prefix}-rotate-@{i}";
      .@{z} when (mod(@i, 90) = 0) {
        &:extend(.ie-rotation-fix);
      }
   }
  .calculateIEFixes(359);
}
```

Both options result in
```css
.ie-rotation-fix, .fa-rotate-0, .fa-rotate-90, .fa-rotate-180, .fa-rotate-270 {
  filter: none;
}
```

This is a bit ugly, especially with the erroneous class; however it seems preferable to the alternative of duplicated declarations.

johnnyshields:
Recommend to keep consistent with Bootstrap-esque naming conventions.

`.fa-xlg`    -->  `.fa-xl`
`.fa-xsm`  -->  `.fa-xs`

zakmac:
Per @johnnyshields suggestion, the following size class names have been modified `.xlg->.xl` and `.xsm->.xs`.

PR #6922, opened 2015-07-10
src/icons.yml: add more filters

williamboman:


src/icons.yml
```diff
@@ -1947,6 +1950,7 @@ icons:
       - checklist
       - todo
       - list
+      - hamburger
```
williamboman: As in a hamburger icon.

tagliala:
thanks!

PR #6950, opened 2015-07-14
Expose palm tree icon

### LINKED ISSUE #3786: Icon request: palm-tree-island
Hi,

Firstly thanks for the icon set!

I'd like to see in icon of a single palm tree on an island, it would be great to represent holiday/vacation destinations.

Thanks,
Ian

gotdibbs:
Fixes #3786

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 #[Font Awesome v4.3.1](http://fontawesome.io)
 ###The iconic font and CSS framework
 
-Font Awesome is a full suite of 519 pictographic icons for easy scalable vector graphics on websites,
+Font Awesome is a full suite of 520 pictographic icons for easy scalable vector graphics on websites,
 created and maintained by [Dave Gandy](http://twitter.com/davegandy).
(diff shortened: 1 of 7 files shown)


tagliala:
Thanks but I cannot accept this PR

You could explain in #3786 how to take advantage of the 1f334 standard unicode point

gotdibbs:
@tagliala No worries. Could you explain why for future reference? Thanks!

tagliala:
@gotdibbs 

for various reasons: 
* those icons are not part of fontawesome
* the standard unicode set is pretty big, so we will pollute docs and dramatically increase css size
* glyphs are not consistent across browsers/platforms
* some system/platforms does not display the above icon (IE8 / XP)

image

(IE8/xp/ietester - IE11/w7 - IE9/w7)

Given that, if you need an icon from the standard unicode set (like `A`, or `%`) this is the best way to do it

gotdibbs:
@tagliala Thanks for the detailed explanation! That all makes perfect sense.

PR #6955, opened 2015-07-15
Replaced protocol-relative URL with HTTPS protocol

drmonkeyninja:
It is no longer considered good practice to use protocol-relative URLs (http://www.paulirish.com/2010/the-protocol-relative-url/). So the CDN request should be made via HTTPs. I've updated the documentation accordingly.


tagliala:
:+1: 

I would definitely merge this. @davegandy ?

PR #7106, opened 2015-07-31
please add another filter icon

amberino:
The current icon i feel could be confusing to some users...especially ones that aren't web/tech savvy. I would like to suggest one like/similar to the one i provided.

custom_android_settings_icon_by_transitoryspace-d5v6vjn


tagliala:
closing here, please open an issue: https://github.com/FortAwesome/Font-Awesome/issues/new

PR #7247, opened 2015-08-17
Add filter heading to header

eelkeblok:
I was searching for this. Seems to be common enough to make it an alias?

### DIFF
--- a/src/icons.yml
+++ b/src/icons.yml
@@ -4428,6 +4428,8 @@ icons:
     id:         header
     unicode:    f1dc
     created:    4.1
+    filter:
+      - heading
     categories:
       - Text Editor Icons
 


tagliala:
could you please change the alias into a filter?

eelkeblok:
Certainly. Is there some sort of guideline what warrants an alias?

tagliala:
aliases add a css class, so you can also use `fa-heading` in your html. Aliases increase css size

Filters only add a search tag to the docs

tagliala:
@eelkeblok perfect, you only need to squash the commits and force push

eelkeblok:
Thanks for the coaching. Should be all set now

tagliala:
perfect, thanks!

It will be live when the docs will be rebuilt

PR #7311, opened 2015-08-26
Add a line to bower.json

cezarpretto:
Add a line to bower.json to be able to inject de css dependency to the project when it is installed via bower.

I really don't know if it's a mistake of you or the new version is that way. If yes you could explain to me why.

Thanks.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,8 @@
   "license": ["OFL-1.1", "MIT", "CC-BY-3.0"],
   "main": [
     "less/font-awesome.less",
-    "scss/font-awesome.scss"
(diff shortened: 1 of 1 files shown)


tagliala:
Unfortunately no mistakes here, please see #6227 and https://github.com/bower/bower.json-spec/commit/0b34d2bb0bdd4fa31fa9728f787f49e170e41f99 

Sorry for the breaking change

tagliala:
Solution: use overrides

http://blog.getbootstrap.com/2015/06/15/bootstrap-3-3-5-released/#wiredep-and-bower

PR #7432, opened 2015-09-10
Synonyms for caret and cc-amex icons

jibbajabba:
This is a request to please add synonyms to be able to search "Triangle" keywords for caret, and "Amex" for American Express Credit Card.


tagliala:
Thanks!

PR #7560, opened 2015-09-28
Created function that returns icons code

somewebmedia:
If anyone wants to use icons without the regular fa classes, but rather applying them to specific elements dynamically.

Example:

```
.upload_button::after {
  @include fa-icon();
  content: fa-icon-var(upload);
}
```

Tested with [node-sass](https://github.com/sass/node-sass) that provides binding for Node.js to [libsass](https://github.com/sass/libsass).


robmadole:

– FA team

PR #7562, opened 2015-09-29
Allow easy use with bundlers

maxdow:
With this new entry in package.json, you can require the font-awesome library easily in bundlers tools like webpack because it resolves directly to the correct file ( like normalize.css do )


maxdow:
up ?

robmadole:

– FA team

PR #7566, opened 2015-09-29
Combine multiple CDN files

jimaek:
Less requests and less points of failure.

[Info on combination](https://github.com/jsdelivr/jsdelivr#load-multiple-files-with-single-http-request)


quantumpacket:
One can argue that it creates a single point of failure by putting more eggs in the same basket. Not to mention now any version update done would cause the request to be new and thus all of those files will download, even ones that were not updated. Keeping the files separate when it comes to version controlled libraries takes advantage of caching.

PR #7607, opened 2015-10-04
Remove warnings on safari

rrgirish:
Safari throws warnings on filter: property which is set for IE. Putting them in quotes removes the warning and leaves it working for IE

### DIFF
--- a/css/font-awesome.css
+++ b/css/font-awesome.css
@@ -118,31 +118,31 @@
   }
 }
 .fa-rotate-90 {
-  filter: progid:DXImageTransform.Microsoft.BasicImage(rotation=1);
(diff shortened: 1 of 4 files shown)


tagliala:
Can you confirm me 100% that this works on IE8 ?

rrgirish:
Don't have IE8.Sorry!
Would you be able to verify this on IE8?

tagliala:
@rrgirish I should be able to do it

xewl:
Actually this kinda states otherwise:
http://lesscss.org/features/#features-overview-feature-escaping

Long story short: The filter's content has to be evaluated as Javascript when compiling.
Fixing this in the .css will not be sufficient. The quotes would work in IE8.

robmadole:

– FA team

PR #7639, opened 2015-10-08
Update package.json

PeterDaveHello:


jisaacks:
According to the `package.json` [docs](https://docs.npmjs.com/files/package.json#license) the current way is closer to what is recommended for multiple licenses.

PeterDaveHello:
Updated!

tagliala:
the operator should be AND, am I right?

* The Font Awesome font is licensed under the SIL OFL 1.1:
http://scripts.sil.org/OFL
* Font Awesome CSS, LESS, and Sass files are licensed under the MIT License:
http://opensource.org/licenses/mit-license.html

jisaacks:
> the operator should be AND, am I right?

I would tend to agree, that is also valid SPDX.

So in that sense, this PR really just needs to add the wrapping parens.

PeterDaveHello:
oops, so I think I should close this, sorry.

tagliala:
@PeterDaveHello I suppose we still need parentheses

PeterDaveHello:
Oh okay

tagliala:
Thanks. Merging because of https://docs.npmjs.com/files/package.json#license

> If your package is licensed under multiple common licenses, use an SPDX license expression syntax version 2.0 string, like this:
```{ "license" : "(ISC OR GPL-3.0)" }```

I'm still worried about breaking backward compatibility but let's follow the specs

PeterDaveHello:
:+1:

Issue #7927, opened 2015-11-19
Add grid filter to icon-th


harobed:
Icon Request : Apps (grid 3x3 square)

harobed:
Like : http://www.englewoodschools.net/cms/lib8/CO01900647/Centricity/Domain/1009/GoogleAppsIcon.png

tagliala:
Does this one suit your use case? http://fontawesome.io/icon/th/

harobed:
@tagliala yes ! I didn't see it in « Text Editor Icons » list. Thanks

tagliala:
I can provide a "grid" filter for this icon

tuxy:
+1 for ![fa-grid](http://www.englewoodschools.net/cms/lib8/CO01900647/Centricity/Domain/1009/GoogleAppsIcon.png)

It should be called **fa-grid**

tagliala:
@tuxy this request is about a doc filter.

gurisko:
@tagliala I suppose we can close this issue.

PR #7951, opened 2015-11-21
Add filter grid

gurisko:
This PR adds `grid` filter to `fa-th` and `fa-th-large` as requested in #7927.


tagliala:
Merged, thanks

PR #7960, opened 2015-11-23
Icon request: square-half-o

bropa:
Hi,

it would be nice to implement an square-half-o icon like star-half-o. We want to use squares instead of stars for reviews.

Thank you for your great work!


tagliala:
@bropa please open a proper request

PR #8222, opened 2015-12-26
fa: add government alias

Joery:
Add fa-government alias to the existing fa-university icon

PR #8226, opened 2015-12-27
fa: add government alias

Joery:
Add fa-government alias to the existing fa-university icon

### DIFF
--- a/src/icons.yml
+++ b/src/icons.yml
@@ -3958,6 +3958,7 @@ icons:
     aliases:
       - institution
       - bank
+      - government
     categories:
       - Web Application Icons
 


tagliala:
https://github.com/FortAwesome/Font-Awesome/pull/8222#issuecomment-167353251

> @Joery please add a filter, not an alias
so you will still need to alias `fa-government` to `fa-university` in your own application but when you search for `government` at http://fontawesome.io/icons/ you will find the icon you need

Joery:
@tagliala
You have a example PR?

src/icons.yml
```diff
@@ -3958,6 +3958,7 @@ icons:
     aliases:
       - institution
       - bank
+      - government
```
tagliala: just add a `filter:` line before this line

filter should be 4 space indented, like `aliases:`

tagliala:
Sure.

See my above comment. Example PR: #7444

PR #8426, opened 2016-01-27
add compress.html layout

marcobiedermann:
This layout will:
- removes unnecessary whitespace
- removes optional end tags
- removes optional start tags
- removes comments
- preserves whitespace within `<pre>`

More information [here](http://jch.penibelst.de/)


robmadole:

– FA team

Issue #4851, opened 2014-10-27
Please include icons.yml file somewhere in the ZIP file that's downloaded


chockenberry:
I'm working on a new feature in [xScope](http://xScopeApp.com) that supports Font Awesome in the Text tool. The feature requires getting a list of the current icons into an internal database, and this can be done easily by importing icons.yml.

The hard part for designers will be to find and download the icons.yml file in the GitHub repository. They'll have to download the entire repository and then dig a single file out of the src folder.

It would be much easier for these users if you added the file to an "info" folder along with the "css", "fonts", "less" and "scss" folders.

dominykas:
Including `icons.yml` in zip or whatever other distribution would also be beneficial for projects like https://github.com/CodeCharmLtd/Font-Awesome-SVG-PNG - they could just depend on the `font-awesome` published in `npm`, without having to explicitly fetch it from github, as described https://github.com/FortAwesome/Font-Awesome/issues/2795

dotherightthing:
I just purchased XScope, but https://raw.githubusercontent.com/FortAwesome/Font-Awesome/master/src/icons.yml is 404. Where is icons.yml now?

tagliala:
@dotherightthing here: https://raw.githubusercontent.com/FortAwesome/Font-Awesome/master/metadata/icons.yml

tagliala:
Done. Zip archive contains the `icons.yml` file in the `/metadata` subfolder

PR #8852, opened 2016-03-29
Exclude icons.yml from .npmignore

HHogg:
**Use case**
For a custom organisation library/framework that has its own documentation (style guide) and using the Sass build to customise the class prefix. It would be nice to generate an icon list (similar to the [official list](http://fontawesome.io/icons/)) that is specific to the custom build. 

I know it could be done just by saying _"Go to the official docs, replace the `fa-` prefix for [CUSTOM PREFIX] and ignore any icons before [PACKAGE VERSION]"_ 

... but I can't see any harm in including the yaml file :smiley_cat: ... plus I'm sure there are other use cases for having access to a full list of the icons. Like in [#4851](https://github.com/FortAwesome/Font-Awesome/issues/4851)

**Note: this will also include the `license.html` and `README.md-nobuild` because they match files that are never ignored**

> The following paths and files are never ignored, so adding them to .npmignore is pointless:
> - package.json
> - README (and its variants)
> - CHANGELOG (and its variants)
> - LICENSE / LICENCE

Taken from [NPM docs](https://docs.npmjs.com/misc/developers)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -32,7 +32,8 @@ stunnel.log
 .ruby-version
 
 # don't need these in the npm package.
-src/
+src/*
+!src/icons.yml
 _config.yml
 bower.json
 component.json


HHogg:
Is there any opinion on this? It's something that I feel would be very useful, with little impact or consequence.

stefancrowe:
+1 for this

It seems there's no agreed way of generating a list of icons for a particular version of Font Awesome.

It would be even more ideal to have a JSON-formatted list of icons included for this purpose (i.e. `icons.json`). However, including the `icons.yml` would still help.

mlazowik:
+1

We need the list of icons in our backend, would be great to generate it automatically based on whatever version of fa is installed instead of managing that manually.

robmadole:

– FA team

Issue #1720, opened 2013-08-04
Icon Request: Marijuana/Cannabis


LDigital84:
Hi,

I would like to request a marijuana/cannabis leaf icon. I know this request could be controversial but I live in Colorado which has both medical and recreational marijuana use. I build websites for the marijuana industry and I absolutely LOVE Font Awesome. If this icon was included it would really make Font Awesome even more Awesome! Also there are no other icon fonts (that I know of) that contain a marijuana leaf.

I would use this on the sites that I build for the marijuana industry. So far I have to rely on images for handling this. Since I already use Font Awesome on many of these sites I would prefer not to have to rely on an image.

I appreciate your consideration.

Lynda 
![marijuana-icon](https://f.cloud.github.com/assets/1938824/907697/5622c408-fd56-11e2-9e54-9b3627a74a29.jpg)

andrewkatz:
+1

G-Driver:
+1

wilblack:
+1
any status change on this?

moesfeld:
+1

xereon:
+1 I've been waiting for the cannabis leaf text/font symbol for years. Font-Awesome is the answer. <3

originalslicey:
+1 There are several icons I need for the websites I build for the medical cannabis community. This is one of them.

wilblack:
+1 I am working on a website for the soon-to-be legal marijuana industry in Oregon. This icon as a font would be Awesome!

Fatbat:
+1 again.

harigopalakrishna:
+1

XenRevo:
+1
+420

mwthink:
+1

Fatbat:
Would really love to see this make it into the next update.

msamoylov:
+1

schulzefelix:
+1

thatrubylove:
still no?

+420

bensayersllc:
+1

420connect:
+100%

beanfan78:
Would be amazing to have this. It would really help on many of my clients sites. +1, the industry is growing (no pun intended), and a font-awesome icon would be another vote of legitimacy!

tbeckman1234:
+1

niteshp:
+1

d1git31:
[...]

PR #8874, opened 2016-04-01
Talbs/add a11y

louizou:
Font awesome lacks only one icon to be the perfect icon set..... 
A real ganja leaf

As the old french philosopher raggasonic said ... 
It must legalize the ganja.... and eliminate the brown resselers :-)    
Sorry for the approximative translation :-)


tagliala:
Please +1 #1720

Issue #6133, opened 2015-04-02
Update accessibility


dylanb:
Please do two things:

1. Remove the mis-representation from your web site until you have properly fixed the problem, so that you do not continue to make the web a worse place for people with disabilities,
2. Fix the accessibility problems properly

karlgroves:
@dylanb It might be good to cite the specific location(s) where the claim is made, this way an interested developer can find & fix the issue and issue a PR

dylanb:
@karlgroves http://fontawesome.io/

karlgroves:
@dylanb For clarity, you're referring to the text in the footer?   Does it exist anywhere else?

Reason I'm asking is I'll be glad to make the necessary edits (or deletions) and issue the PR myself.

karlgroves:
Deleted my prior comment.

@dylanb - as an issue description, I think you know it should have better information in it so that action could be taken

@davegandy - The claim made in the footer of the http://fontawesome.io/ website which says " Screen Reader Compatible" is demonstrably false.

Given the icon fa-bed (http://fontawesome.io/icon/bed/)

Using the following markup: <i class="fa fa-bed"></i>

It creates CSS generated content: "\f236";

There are three accessibility problems that may result:

1. (and this is @dylanb's point) Screen readers *may* read that CSS generated content.
    The workaround for this scenario is to use aria-hidden like so: <i class="fa fa-bed" aria-hidden="true"></i> which would allow a screen reader to ignore it. BUT this raises another concern

2. Because it creates generated content, there's no text alternative.  
    Imagine I did this <a href="http://www.github.com"><i class="fa fa-github"></i></a>
    That has no content in the text node of the DOM. It is, in almost all cases, a blank link. Going back to the above, users *may* get access to the generated content which, in all likelihood, actually sucks to hear on a screen reader.  So here's the fix

    <a href="http://www.github.com"><i class="fa fa-github" aria-hidden="true"></i><span class="sr-only">Github</span></a>
 
    That uses the 'sr-only' class naming convention from Bootstrap, BTW. That's not necessary, but what is necessary is basically that (if you don't want the word "Github" to appear) the span be positioned off screen to hide it.  This is effectively the same as having an alt attribute on an image

    But wait! There's more

[...]

PR #8879, opened 2016-04-02
Add Accessibility Information and Best Practices to Doc Site

talbs:
**This work aims to tackle the following:**
* Adds in an informational page on Font Awesome (and icon fonts) and web accessibility - initially including background, use cases, and more information.
* Adds accessibility-minded examples to the examples page
* Updates the markup displayed on an icon detail page to use `aria-hidden="true"` with a link to more information on the new accessibility page
* Generally updates the Font Awesome site with some of these best practices.

**What this doesn't cover**
* adding or using [advanced icon fallback techniques](https://www.filamentgroup.com/lab/bulletproof_icon_fonts.html) for high contrast mode
* Updates to test-based files that are not end user facing

- - -

#### Reviewers

* [x] Copy and UI - @davegandy
* [ ] a11y - @dylanb

- - -

FYI, @tagliala. I'll comment on the few open GH issues that I'm hoping this work addresses as we progress. With the PR process in mind, would you recommend asking those members to review/take a look at this work?

### DIFF
--- a/src/_includes/accessibility/background.html
+++ b/src/_includes/accessibility/background.html
@@ -1,0 +1,15 @@
+<div id="background">
+  <h2 class="page-header">Icon Fonts &amp; Accessibility</h2>
+  <p>
+    Modern versions of assistive technology, like screen readers, will read CSS generated content (how Font Awesome icons are rendered), as well as specific Unicode characters. When trying our default markup for rendering icons, assisistive technology may have the following problems.
+  </p>
+
+  <ul>
+    <li>
+      The assistive technology may not find any content to read out to a user
+    </li>
+    <li>
+      The assistive technolog may read the unicode equivalent which could not match up to what the icon means in context or worse is just plain confusing
+    </li>
+  </ul>
+</div>
--- a/src/_includes/accessibility/using-with-accessibility.html
+++ b/src/_includes/accessibility/using-with-accessibility.html
@@ -1,0 +1,130 @@
+<div id="using-with-accessibility">
+  <h2 class="page-header">Using Font Awesome with Acessibility in mind</h2>
+  <p>
+    When using icons in your UI, there are ways to help assistive technology either ignore or better understand Font Awesome.
+  </p>
+
+  <h3>Icons used for pure decoration or visual styling</h3>
+  <p>
+    If you're using an icon to add some extra decoration or branding, it does not need to be announced to users as they are navigating your site or app aurally. Additionally, if you're using an icon to visually re-emphasize or add styling to content already present in your HTML, it does not need to be repeated to an assistive technology-using user. You can make sure this is not read by adding the <code>aria-hidden="true"</code> to your Font Awesome markup.
+  </p>
+
+  <div class="margin-top-lg margin-bottom-lg">
+{% highlight html %}
+<i class="fa fa-fighter-jet" aria-hidden="true"></i>
+{% endhighlight %}
+<small class="text-muted">an icon being used as pure decoration</small>
+  </div>
+
+  <div class="margin-bottom-lg">
+{% highlight html %}
+<h1 class="logo">
+  <i class="fa fa-pied-piper" aria-hidden="true"></i>
+  Pied Piper, A Middle-Out Compression Solution Making Data Storage Problems Smaller
+</h1>
+{% endhighlight %}
+<small class="text-muted">an icon being used as a logo</small>
+  </div>
+
+  <div class="margin-bottom-lg">
+{% highlight html %}
+<a href="https://github.com/FortAwesome/Font-Awesome"><i class="fa fa-github" aria-hidden="true"></i> View this project's code on Github</a>
+{% endhighlight %}
+<small class="text-muted">an icon being used in front of link text</small>
+  </div>
+
+  <h3>Icons that with semantic or interactive purpose</h3>
+  <p>
+    If you're using an icon to convey meaning (rather than only as a decorative element), ensure that this meaning is also conveyed to assistive technologies. This goes for content you're abbreviating via icons as well as interactive controls (buttons, form elements, toggles, etc.). There are a few techniques to accomplish this:
+  </p>
+
+  <h4>If an icon is <strong>not</strong> an interactive element</h4>
+  <p>
+    If not representing an interactive element, like a button or link, use both a <code>title</code> attribute and an  on the icon to provide a text alternative. This has the advantages of working with all modern screen readers, working well with the most ubiquitous speech recognition program, and helping users with cognitive disabilities by providing a mouse tooltip.
+  </p>
+
+  <div class="margin-bottom-lg margin-top-lg">
+{% highlight html %}
+<dl>
+  <dt>
+    <i class="fa fa-car" aria-hidden="true"></i>
+    <span class="sr-only">Time to destination by car:</span>
+  </dt>
+  <dd>4 minutes</dd>
+
+  <dt>
+    <i class="fa fa-bicycle" aria-hidden="true"></i>
+    <span class="sr-only">Time to destination by bike:</span>
+  </dt>
+  <dd>12 minutes</dd>
+</dl>
+{% endhighlight %}
+    <small class="text-muted">an icon being used to communicate travel methods</small>
+  </div>
+
+  <div class="margin-bottom-lg margin-top-lg">
+{% highlight html %}
+<i class="fa fa-hourglass" aria-hidden="true"></i>
+<span class="sr-only">60 minutes remains in your exam</span>
+
+<i class="fa fa-hourglass-half" aria-hidden="true"></i>
+<span class="sr-only">30 minutes remains in your exam</span>
+
+<i class="fa fa-hourglass-end" aria-hidden="true"></i>
+<span class="sr-only">0 minutes remains in your exam</span>
+{% endhighlight %}
+    <small class="text-muted">an icon being used to denote time remaining</small>
+  </div>
+
+  <h4>If an icon represents an interactive element</h4>
+  <p>
+    In addition to the recommendations above, use a <code>title</code> attribute which makes the accessible text discoverable by mouse and those with cognitive disabilities.
+  </p>
+
+  <div class="margin-bottom-lg margin-top-lg">
+{% highlight html %}
+<a href="path/to/shopping/cart">
+  <i class="fa fa-shopping-cart" title="View 3 items in your shopping cart" aria-hidden="true"></i>
+  <span class="sr-only">View 3 items in your shopping cart</span>
+</a>
+{% endhighlight %}
+    <small class="text-muted">an icon being used to communicate shopping cart state</small>
+  </div>
+
+  <div class="margin-bottom-lg margin-top-lg">
+{% highlight html %}
+<a href="#navigation-main">
+  <i class="fa fa-bars" title="Skip to main navigation" aria-hidden="true"></i>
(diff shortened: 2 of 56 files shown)


tagliala:
@talbs thanks. The most important one should be #6133

src/_includes/examples/accessible.html
```diff
@@ -0,0 +1,96 @@
+<section id="accessible">
+  <h2 class="page-header">
+    Accessibility-Minded
+  </h2>
+  <div class="row">
+    <div class="col-md-3 col-sm-4">
+      <p>
+        <a class="btn btn-default" href="path/to/settings">
+          <i class="fa fa-cog" aria-hidden="true"><span class="sr-only">Settings</span></i>
```
dylanb: adding `sr-only` to something whose parent is has `aria-hidden="true"` will render it invisible to screen readers. You can fix this in two ways. Either (preferred)
```
<a class="btn btn-default" href="path/to/settings" title="Settings">
          <i class="fa fa-cog" role="presentation"></i>
</a>
```
or (lots of redundancy)
```
<a class="btn btn-default" href="path/to/settings" title="Settings">
          <i class="fa fa-cog" aria-hidden="true"><span class="sr-only">Settings</span></i>
</a>
```
This will also ensure that the anchor will work for systems like Dragon

src/_layouts/icon.html
```diff
@@ -49,10 +49,15 @@ <h1 class="info-class">
           </div>
         </div>
 {% highlight html %}
-<i class="fa fa-{{ page.icon.id }}"></i>
+<i class="fa fa-{{ page.icon.id }}" aria-hidden="true"></i>
 {% endhighlight %}
-        <br>
-        <div class="lead"><i class="fa fa-info-sign"></i> Looking for more? Check out the <a href="{{ page.relative_path }}examples/">examples</a>.</div>
+        <p class="margin-bottom-lg">
```
dylanb: role="presentation" is more semantically correct that aria-hidden="true"

src/_includes/brand-adblock-warning.html
```diff
@@ -1,4 +1,4 @@
-<h4><i class="fa fa-warning"></i> Warning!</h4>
+<h4><i class="fa fa-warning" aria-hidden="true"></i> Warning!</h4>
```
dylanb: see previous comment re: role="presentation"

src/_includes/accessibility/using-with-accessibility.html
```diff
@@ -0,0 +1,78 @@
+<div id="using-with-accessibility">
+  <h2 class="page-header">Using Font Awesome with Acessibility in mind</h2>
+  <p>
+    When using icons in your UI, there are ways to help assistive technology either ignore or better understand Font Awesome.
+  </p>
+
+  <h3>Icons used for pure decoration or visual styling</h3>
+  <p>
+    If you're using an icon to add some extra decoration or branding, it does not need to be announced to users as they are navigating your site or app aurally. Additionally, if you're using an icon to visually re-emphasize or add styling to content already present in your HTML, it does not need to be repeated to an assistive technology-using user. You can make sure this is not read by adding the <code>aria-hidden="true"</code> attribute to your Font Awesome markup.
```
dylanb: cf. previous comment re: role="presentation"

src/_includes/accessibility/using-with-accessibility.html
```diff
@@ -0,0 +1,78 @@
...
+  <div class="margin-bottom-lg">
+{% highlight html %}
+<a href="https://github.com/FortAwesome/Font-Awesome"><i class="fa fa-github" aria-hidden="true"></i> View this project's code on Github</a>
+{% endhighlight %}
+<small>an icon being used in front of link text</small>
+  </div>
+
+  <h3>Icons that with semantic or interactive purpose</h3>
+  <p>
+    If you're using an icon to convey meaning (rather than only as a decorative element), ensure that this meaning is also conveyed to assistive technologies – for instance, include additional content that's visually hidden to users but semantically avaiable to screen readers. This goes for content you're abbreviating via icons as well as interactive controls (buttons, form elements, toggles, etc.).
```
dylanb: it would be better to use the title attribute because of the following advantages

1. It works with all modern screen readers
2. It works well with the most ubiquitous speech recognition program (Dragon)
3. It help users with cognitive disabilities by providing a mouse tooltip

talbs:
FYI, @cptvitamin. Here's the PR I mentioned to bring Font Awesome's info and examples up to speed with good a11y practices. Mind taking a look?

src/_includes/accessibility/background.html
```diff
@@ -0,0 +1,15 @@
+<div id="background">
+  <h2 class="page-header">Icon Fonts &amp; Accessibility</h2>
+  <p>
+    Modern versions of assistive technology, like screen readers, will read CSS generated content (how Font Awesome icons are rendered), as well as specific Unicode characters. When trying our default markup for rendering icons, assisistive technology may have the following problems.
+  </p>
+
+  <ul>
+    <li>
+      The assistive technology may not finding any content to read out to a user
```
cptvitamin: finding -> find

talbs:
@dylanb, can you walk me through why and exactly when would using a title attribute be better to recommend than the `<span class="sr-only">alternative text equivalent</span>`?

I want to make sure we have the right syntax for the right cases in our examples. Is there ever a case you'd want to use  `<span class="sr-only">alternative text equivalent</span>` rather than the title attribute suggestion you had?

src/_layouts/icon.html
talbs: Should be resolved.

src/_includes/brand-adblock-warning.html
talbs: Should be resolved.

talbs:
@dylanb, I think I addressed all of your feedback. Mind taking another look as well as chiming in on my questions above (https://github.com/FortAwesome/Font-Awesome/pull/8879#issuecomment-205370942)?

dylanb:
@talbs Definitely looks much better for screen readers

talbs:
> Apparently role="presentation" is wrong (#6133 (comment)) I would go only for aria-hidden="true" as before

Thanks, @tagliala. I'll revert back to using the `aria-hidden="true"` technique rather than the `role="presentation"` route.

talbs:
@dylanb, thanks for the continued help. Mind taking a hopefully final look and giving this work your :+1:  if you think its ready for primetime?

src/_includes/community/requesting-new-icons.html
```diff
@@ -23,7 +23,7 @@ <h2 class="page-header">Requesting New Icons</h2>
       </ul>
     </li>
     <li>
-      Request concrete objects: it's harder to make an icon to represent happiness, it's easier to make a smiley face.  <i class="fa fa-smile-o"></i>
+      Request concrete objects: it's harder to make an icon to represent happiness, it's easier to make a smiley face.  <i class="fa fa-smile-o" aria-hidden="true"></i>
```
dylanb: I would argue that this example is not purely decorative and therefore needs an alt text

src/_includes/examples/animated.html
```diff
@@ -9,11 +9,11 @@ <h2 class="page-header">
   <div class="row">
     <div class="col-md-3 col-sm-4">
       <p>
-        <i class="fa fa-spinner fa-spin fa-3x fa-fw margin-bottom"></i>
-        <i class="fa fa-circle-o-notch fa-spin fa-3x fa-fw margin-bottom"></i>
-        <i class="fa fa-refresh fa-spin fa-3x fa-fw margin-bottom"></i>
-        <i class="fa fa-cog fa-spin fa-3x fa-fw margin-bottom"></i>
-        <i class="fa fa-spinner fa-pulse fa-3x fa-fw margin-bottom"></i>
+        <i class="fa fa-spinner fa-spin fa-3x fa-fw margin-bottom" aria-hidden="true"></i>
```
dylanb: This list of icons is communicating that something is busy, they should have alt texts

talbs:
@dylanb, your suggestion to use `role="presentation"` for any icons that are purely decorative is really appealing, since its semantic and more straight-forward. **Can you confirm that `role="presentation"` is enough to hide all CSS-generated content (we use the `:before` and `:after` pseudo-elements to help render our icons) from assistive technology?**

If that's solid enough, we can rely on that in general rather than `aria-hidden="true"`.

dylanb:
I apologize for the confusion: role="presentation" is not going to hide content. When I made that suggestion, I had not yet internalized the desire to hide the unicode character and was interpreting this as a desire to hide any semantics on the `<i>` tag.

To hide content from screen readers, you must use `aria-hidden="true"`

dylanb:
ok, looking through your most recent changes, these have made things a mixture of better and worse and its probably my fault for not explaining adequately. I have created three bullet points below to summarize the salient information. If you want me to turn this into examples, please let me know.

1. `aria-hidden="true"` will hide all content and semantics and is the technique that should be used to hide unwanted unicode characters
1. `title` attribute should only be used as an accessible name on interactive elements like links or buttons, in all other cases, `sr-only` should be used in addition to title to support users with cognitive disabilities (see additional information below)
1. role="presentation" could be used on the `<i>` tags in cases where the unicode character is meaningful. This is really a "futureproofing" best practice because most screen reader settings do not cause the semantics of that tag to be exposed (but they could).

Additional Information: the title attribute is part of the ARIA name calculation algorithm as the "fallback". ARIA name calculation, while well supported for well known role combinations, is not supported by any browser+AT combinations in its entirety. So for example, using aria-label on a `<div>` will not work on most platforms, while using it on an interactive element like a button or a link, will work everywhere. The same applies to the title attribute. Therefore, title will provide the most bang for the buck when used on an interactive element, where it can serve as the accessible name, the visible hint and the target for Dragon. When used on non-interactive elements, it is really only serving the purpose of a hint which helps users with cognitive disabilities.

talbs:
>
1. `aria-hidden="true"` will hide all content and semantics and is the technique that should be used to hide unwanted unicode characters
1. `title` attribute should only be used as an accessible name on interactive elements like links or buttons, in all other cases, `sr-only` should be used in addition to `title` to support users with cognitive disabilities (see additional information below)
1. role="presentation" could be used on the `<i>` tags in cases where the unicode character is meaningful. This is really a "futureproofing" best practice because most screen reader settings do not cause the semantics of that tag to be exposed (but they could).

Perfect! Thanks, @dylanb. That's exactly the feedback I was looking for. I'll sync up our examples, sample icon code, and doc site markup to represent that. Expect 1 more round of spam shortly, :)

talbs:
A bit of status, all.

I've updated the examples, a11y page, and the Font Awesome site's markup to reflect the most recent use cases of icons and accessibility per @dylanb. Also, all commits have been squashed.

We've also decided to roll this work into a version bump of Font Awesome, which has a few more a11y-based goodies. With that said, once the dust settles here, I'll close this thread and open a PR to merge this into that version's branch.

@dylanb and @davegandy, is this good to go from your perspectives?

talbs:
Since this seems solid from both a11y and doc/UI points of view, I'm closing this PR as this work is being rolled into the [WIP 4.6.0 branch focusing in accessibility](https://github.com/FortAwesome/Font-Awesome/commits/4.6.0-wip).

Thanks for all of the help, folks. Really excited to get these improvements out to our users. :)

talbs:
Also, I'll leave this branch open until the `4.6.0` version goes live. **If you have any further feedback, please provide it here and I'll address it.**

Issue #3614, opened 2014-06-03
Icon Request: icon-wine-glass-full


singlewhitepixel:
I recently required a wine glass icon, two versions one full and one empty. I used an image but you may want to consider it and other glass shapes as I often need alcoholic beverage related icons for bars, liquor retailers and restaurant related websites and I'm sure other web designers would too.

G-Driver:
+1 for wine

jaoustin:
+1

phillq:
+1

G-Driver:
See also corkscrew request https://github.com/FortAwesome/Font-Awesome/issues/5835

kewh:
+1

awhare:
+1

jkfoto:
+1

bigsea:
+1

willythekid1979:
+1

playcodestudio:
+1

Crissov:
🍷 [U+1F377](http://emojipedia.org/wine-glass/) Unicode emoji #222

clabnet:
+1

vshvedov:
+1

tzeller:
i have created this icon - no copy or so - which can be straight used. i need as well some more for the food industry which i will attach
[fa-wine.pdf](https://github.com/FortAwesome/Font-Awesome/files/1364626/fa-wine.pdf)

tzeller:
wine glass 14 px and 32px
![fa-wine_14px](https://user-images.githubusercontent.com/859700/31308528-7afbf32e-ab78-11e7-8bb1-9fc7b3447585.png)
![fa-wine_32px](https://user-images.githubusercontent.com/859700/31308531-7d9d6ba8-ab78-11e7-8825-735fc156fbe9.png)

carterdesign:
+1

Monster695:
+1

sensibleworld:
We have a few options for this (with the inclusion of Font Awesome Pro):

https://fontawesome.com/icons/wine-glass?style=solid
https://fontawesome.com/icons/wine-glass-alt?style=solid
https://fontawesome.com/icons/wine-glass?style=regular

Issue #370, opened 2012-07-30
Icon request: icon-water-glass, icon-wine-glass, icon-whisky-glass


SweVictor:
Hi,
Right now there is a very nice drink glass icon (icon-glass). It would be very nice if there were some other glasses as well (water, wine, whisky)

Thanks for a great project!
/Victor

emilklindt:
+1 on water

G-Driver:
+1

<img src="http://static.vecteezy.com/system/resources/previews/000/044/420/original/wine%20glass%20symbol.jpg">

phillq:
+1

kewh:
+1

w5d:
+1

shanetheboland:
+1

bigsea:
+1 wineglass!!

lori:
+1 wine glass

kaiazeekee:
+1 for glassware that isn't necessarily alcohol related (working with charity events)

Possibly ice tea class with a lemon? (also from the south)

santinoalbert:
+1 for wine icons!

JJ69:
+1 for wineglass icon

willythekid1979:
+1

revweb:
+1

nbunney:
+1 for all three

nathancouch:
Any update on wine glass and bottle icons?

carolaclavo:
+1

Melloyell:
+1

Crissov:
## Unicode Glass Emojis
#222 

- 🍺 [U+1F37A](http://emojipedia.org/beer-mug/) and 🍻 [U+1F37B](http://emojipedia.org/clinking-beer-mugs/) Beer Mugs
- 🍷 [U+1F377](http://emojipedia.org/wine-galss/) Wine Glass
- 🍸 [U+1F378](http://emojipedia.org/cocktail-glass/) Cocktail Glass
- 🍹 [U+1F379](http://emojipedia.org/tropical-drink/) Tropical Drink 
- 🍾 [U+1F37E](http://emojipedia.org/bottle-with-popping-cork/) Champagne Bottle
- 🍼 [U+1F37C](http://emojipedia.org/baby-bottle/) Baby Bottle
- 🥃 [U+1F943](http://emojipedia.org/tumbler-glass/) Tumbler/Whiskey Glass
[...]

PR #9036, opened 2016-04-21
Adding 'wine' search filter to glass icon

bengadbois:
The alcoholics may disagree, but martini and wine glasses are visually similar enough that it may be good additional filter


tagliala:
Sorry, I'm going to close this one.

We have a couple of requests for wine glasses #3614, #370

PR #9043, opened 2016-04-22
Multiple changes in CDN links

ricardopolo:
- Moving the resources from CDNJs to JSDelivr. In HTTP/2 and SPDY it is recommended to host all dependencies in the same CDN or web server, so all resources are downloaded using the same HTTP connection. It's faster than establish a connection to two CDN providers. In the past using multiple CDN was the best option, but no longer. So we have to put them in just one CDN, I don't have any relation with JSDelivr but think is the best available option.
- Added SRI (integrity) to external scripts. This prevents an infection if the CDN is compromised.
- I removed version aliasing from Algolia (3) and put specific version (3.13.1). This way the cache TTL is longer and avoid breaking changes if the library is updated.
- I added the JSDelivr RUM script. This allow them to test the performance of downloading scripts and takes better load balancing decision for the CDN resources.

### DIFF
--- a/src/_layouts/base.html
+++ b/src/_layouts/base.html
@@ -59,10 +59,11 @@
   <script src="https://ajax.googleapis.com/ajax/libs/jquery/{{ site.jquery.version }}/jquery.min.js"></script>
   <script src="https://ajax.aspnetcdn.com/ajax/jquery.validate/{{ site.jquery_validate.version }}/jquery.validate.min.js"></script>
   <script src="https://maxcdn.bootstrapcdn.com/bootstrap/{{ site.bootstrap.version }}/js/bootstrap.min.js"></script>
-  <script src="https://cdn.jsdelivr.net/algoliasearch/3/algoliasearch.min.js"></script>
-  <script src="https://cdn.jsdelivr.net/algoliasearch.helper/2/algoliasearch.helper.min.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore-min.js"></script>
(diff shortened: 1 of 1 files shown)


ricardopolo:
cc @jimaek

tagliala:
:+1: 

Could you please add your post to the commit message?

@davegandy @talbs can I merge?

ricardopolo:
I dont know how to edit the message of a past commit @tagliala

tagliala:
Amend last commit with:
```bash
$ git commit --amend
```

Edit the commit message:
```
Multiple changes in CDN links

* Moving the resources from CDNJs to JSDelivr. In HTTP/2 and SPDY it is recommended to host all dependencies in the same CDN or web server, so all resources are downloaded using the same HTTP connection. It's faster than establish a connection to two CDN providers. In the past using multiple CDN was the best option, but no longer. So we have to put them in just one CDN, I don't have any relation with JSDelivr but think is the best available option.
* Added SRI (integrity) to external scripts. This prevents an infection if the CDN is compromised.
* Removed version aliasing from Algolia (3) and put specific version (3.13.1). This way the cache TTL is longer and avoid breaking changes if the library is updated.
* Added the JSDelivr RUM script. This allow them to test the performance of downloading scripts and takes better load balancing decision for the CDN resources.
```

then push force with:
```bash
$ git push -f
```

ricardopolo:
@tagliala done

tagliala:
@ricardopolo thanks, it's perfect now. I will wait feedback from Dave or Brian before merging this

davegandy:
Seems fine. If it blows something up, I'll back it out.

Issue #9041, opened 2016-04-22
Improve icon search on the official site


alvarotrigo:
Right now the provided search box by Angolia only looks for the query in the name of the icon.
This usually doesn't result in very useful results as it is basically the same as doing a CTRL+F in the site.

Something far more useful would be adding a couple of keyboards to each icon so users can easily find the icon they are looking for.

A good example of this can be found on [glyphsearch](http://glyphsearch.com/?library=font-awesome) where one can easily find the `fa-smile-o` by looking for the term *face*, for example.

Right now glyphsearch doesn't seem to have all fontawesome icons and I guess it won't get updated at the same time as fontawesome does. 
So seeing this in the official website would be a huge bonus to me, and I guess, for many others who sometimes spend minutes looking through all the 600+ icons just to find the one we are looking for.

PeterShaggyNoble:
That's how it works already. For example, if you search for "menu", fa-bars will be returned as "menu" is a keyword for that icon. You'll also see "menu" in brackets after the icon name.

You can see the full list of keywords for each icon in the [icons.yml file](https://github.com/FortAwesome/Font-Awesome/blob/master/src/icons.yml).

alvarotrigo:
Then I think more keyboards need to be added in some cases. 
For example *money* shows [14 results on glyphsearch](http://glyphsearch.com/?library=font-awesome&query=money) vs 3 of the official search

tagliala:
Feel free to submit a PR with more filters

tagliala:
Closing here, hope it helps

PR #9065, opened 2016-04-26
Adding a few more filters as per #9041

alvarotrigo:
Adding a few more filters as per #9041 
Mostly taken from http://glyphsearch.com/?library=font-awesome


tagliala:
Thanks!

PR #9162, opened 2016-05-10
avoid punctuation in some links

Haroenv:
For example on npm the link would be made with the colon in there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ created and maintained by [Dave Gandy](https://twitter.com/davegandy).
 Stay up to date with the latest release and announcements on Twitter:
 [@fontawesome](http://twitter.com/fontawesome).
 
-Get started at http://fontawesome.io!
+Get started at <http://fontawesome.io>!
 
 ## License
 - The Font Awesome font is licensed under the SIL OFL 1.1:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,6 +98,6 @@ Build the project and documentation:
 
     $ bundle exec jekyll build
 
-Or serve it on a local server on http://localhost:7998/Font-Awesome/:
+Or serve it on a local server on >http://localhost:7998/Font-Awesome/>:
```
tagliala: is `>http://localhost:7998/Font-Awesome/>` right?

Maybe it should be `<http://localhost:7998/Font-Awesome/>`

but we really don't want a link to localhost, maybe `` `http://localhost:7998/Font-Awesome` `` is better

tagliala:
Thanks, but this file is generated, please fix the proper readme: https://github.com/FortAwesome/Font-Awesome/blob/master/src/README.md-nobuild

Please also take a look at my comment

README.md
Haroenv: Yeah, that's a typo. But I think a link to local host still would be beneficial

tagliala: I'm strongly against broken links in the docs

Haroenv:
I'll take a look at using the other file later today

Haroenv:
Solved by #11502

Issue #8368, opened 2016-01-19
'main' field in package.json


fritx:
See: Make normalize.css available for webpack
https://github.com/necolas/normalize.css/pull/421

tagliala:
Sorry, I didn't get this request

Could you please clarify?

fritx:
OK. I meant the `main` field is missing in package.json.
I can't `import` it with webpack. The issue is similar with necolas/normalize.css#421.
Any pull request is welcomed?

tagliala:
Sorry but I'm not so much into node and npm

I'm noticing that `main` seems related to `.js` files

https://docs.npmjs.com/files/package.json#main

> The main field is a module ID that is the primary entry point to your program. That is, if your package is named foo, and a user installs it, and then does require("foo"), then your main module's exports object will be returned.

> This should be a module ID relative to the root of your package folder.

> For most modules, it makes the most sense to have a main script and often not much else.

EG: bootstrap https://github.com/twbs/bootstrap/blob/master/package.json#L22

since there is no javascript here, what is the use case of pointing main to a css file?

fritx:
hi @tagliala
npm is not only for js now. many css and other modules are available on it, such as normalize.css, 
and we prefer to require them in js thanks to webpack bundling.

tagliala:
hi @fritx,

thanks for your clarification. I need feedback from Dave before adding this

fritx:
cool awesome :)

romiem:
+1 Please add a main property in the package.json file which points to ./css/font-awesome.css'.

Youmoo:
+1 [bulma](https://github.com/jgthms/bulma/blob/master/package.json#L7) also does this.

tothandras:
@tagliala sorry for including you but I see this is open for a long time and it would be a quick fix :) I am happy to open a PR too if it makes easier.

tagliala:
@tothandras sorry I can't make this decision, please check if you can feedback from him

niftylettuce:
Please add this!

jrharshath:
Update/FYI: Bootstrap has not only JS, but also CSS. They're using a "style" field in package.json for pointing to css (or scss, in case of bootstrap-sass), and the "main" field for their JS.

fritx:
@jrharshath is this issue related? https://github.com/webpack/css-loader/issues/51
If using the style field, i'm not sure, does `import 'font-awesome'` work? 😕

[...]

PR #9430, opened 2016-06-21
Add main field to package.json

Qard:
This allows support for `npm install font-awesome` and `@import "font-awesome"` to just work
when using a custom sass importer like [node-sass-import](https://www.npmjs.com/package/node-sass-import).

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "description": "The iconic font and CSS framework",
   "version": "4.6.3",
   "style": "css/font-awesome.css",
+  "main": "scss/font-awesome.scss",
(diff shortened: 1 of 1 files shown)


tagliala:
#8368 asks to add the .css file

Qard:
Seems a user in there mentioned Bulma using the main field, but it actually uses it to point at the scss file, like this PR, not the css file.

https://github.com/jgthms/bulma/blob/master/package.json#L7

Not too sure about the webpack use case, as I don't use it myself.

robmadole:

– FA team

PR #9502, opened 2016-06-30
Additional filters so all the comment icons are consistent

bengadbois:
Also, the order is now consistent between them all (hence the rearranging)


tagliala:
Merged, thanks!

Issue #9566, opened 2016-07-13
Accessibility of examples for interactive icons 


adildostmohamed:
According to the post on how to use icons accessibly ( http://fontawesome.io/accessibility/) for icons that convey interaction and therefore should be consumed by a screen reader, the suggested example makes use of an anchor tag with a title attribute that wraps the icon that is set to aria-hidden. The use of a title attribute on an anchor element is considered to be a challenge for many user agents and should be used with caution (as per http://www.w3.org/TR/WCAG20-TECHS/H33.html). In this instance a more accessible example maybe to include a span within the anchor tag that is accessible only to a screen reader with content to provide link purpose (as is used in other examples on the font awesome site) or if this is not feasible to use the aria-label attribute rather than the title attribute as a means of proving link definition to screen reader users.

tagliala:
Thanks for reporting this. I'm asking for feedback from folks at #6133

tagliala:
Examples at http://fontawesome.io/examples/#accessible seem ok

```html
<a class="btn btn-default" href="path/to/settings">
  <i class="fa fa-cog" title="Settings" aria-hidden="true"></i>
  <span class="sr-only">Settings</span>
</a>
```

adildostmohamed:
no problem, for clarity my suggestion is to update the examples used for "If an icon represents an interactive element" from:
`<a href="path/to/shopping/cart" title="View 3 items in your shopping cart">
  <i class="fa fa-shopping-cart" aria-hidden="true"></i>
</a>`

to either:

``<a href="path/to/shopping/cart">
  <i class="fa fa-shopping-cart" aria-hidden="true"></i>
 <span class="sr-only">Shopping cart</span>
</a>``

or

`<a href="path/to/shopping/cart" aria-label="Shopping cart">
  <i class="fa fa-shopping-cart" aria-hidden="true"></i>
</a>`

adildostmohamed:
Sorry just saw your last comment, the examples used there are correct all i would suggest is perhaps updating the code snippets provided at http://fontawesome.io/accessibility/ to ensure that anyone using the code examples are consistent with your examples elsewhere

karlgroves:
I agree with the above. The `title` attribute is less desirable than `aria-label`, especially in the case that @adildostmohamed mentioned.

tagliala:
Is it fine to have both `title` and `aria-label` ?

[...]

PR #9586, opened 2016-07-18
Update Interactive Element Accessibility Documentation

talbs:
This work updates the interactive element portion of the accessibility documentation contained with Font Awesome. In short, it moves to using the `aria-label`-based approach suggested in #9566. All explanations, backgrounds, as well as examples should be updated.

Any cases where an interactive element was being described, we've moved to using the following syntax and recommendation of using an `aria-label` on the interactive element itself (vs. the `title` or `.sr-only` approach).

**Example of New Syntax**
```html
<a href="path/to/shopping/cart" class="btn btn-primary" aria-label="View 3 items in your shopping cart"
  <i class="fa fa-shopping-cart" aria-hidden="true"></i>
</a>
```

**Note**: For non-interactive elements, we continue to recommend the `.sr-only` approach.

- - -

##### Reviewers

* [x] @tagliala 
* [ ] @karlgroves
* [x] @adildostmohamed

### DIFF
--- a/src/_includes/accessibility/accessibility-manual.html
+++ b/src/_includes/accessibility/accessibility-manual.html
@@ -79,7 +79,7 @@
 
   <h4 class="margin-top-xl text-muted">If an icon represents an interactive element</h4>
   <p>
-    In the case of focusable interactive elements, there are various options to include an alternative text or label to the element, without the need for any visually hidden <code>&lt;span&gt;</code> or similar. For instance, simply adding the <code>aria-label</code> attribute with a text description to the interactive element itself will be sufficient to provide an accessible alternative name for the element, as well as providing the mouse tooltip.
+    In the case of focusable interactive elements, there are various options to include an alternative text or label to the element, without the need for any visually hidden <code>&lt;span&gt;</code> or similar. For instance, simply adding the <code>aria-label</code> attribute with a text description to the interactive element itself will be sufficient to provide an accessible alternative name for the element. If you need to provide a visual tooltip on mouseover/focus, we recommend additionally using the <code>title</code> attribute or a <a href="https://github.com/chinchang/hint.css">custom tooltip</a> solution.
   </p>
 
   <div class="margin-bottom-lg margin-top-lg">
@@ -98,10 +98,12 @@
 </a>
 {% endhighlight %}
     <small class="text-muted">an icon being used as a link to a navigation menu</small>
+  </div>
 
+  <div class="margin-bottom-lg margin-top-lg">
 {% highlight html %}
 <a class="btn btn-danger" href="path/to/settings" aria-label="Delete">
-  <i class="fa fa-trash-o" aria-hidden="true" title="Delete"></i>
+  <i class="fa fa-trash-o" aria-hidden="true" title="Delete this item?"></i>
 </a>
 {% endhighlight %}
     <small class="text-muted">an icon being used as a delete button's symbol with a <code>title</code> attribute to provide a native mouse tooltip</small>
--- a/src/_includes/accessibility/accessibility-manual.html
+++ b/src/_includes/accessibility/accessibility-manual.html
@@ -79,12 +79,12 @@
 
   <h4 class="margin-top-xl text-muted">If an icon represents an interactive element</h4>
   <p>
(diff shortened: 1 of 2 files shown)


src/_includes/accessibility/accessibility-manual.html
```diff
@@ -79,12 +79,12 @@ <h4 class="margin-top-xl text-muted">If an icon is <strong>not</strong> an inter
 
   <h4 class="margin-top-xl text-muted">If an icon represents an interactive element</h4>
   <p>
-    In the case of focusable interactive elements, there are various options to include an alternative text or label to the element, without the need for any visually hidden <code>&lt;span&gt;</code> or similar. For instance, simply adding the <code>title</code> attribute to the interactive element itself will be sufficient to provide an accessible alternative name for the element, as well as providing the mouse tooltip.
+    In the case of focusable interactive elements, there are various options to include an alternative text or label to the element, without the need for any visually hidden <code>&lt;span&gt;</code> or similar. For instance, simply adding the <code>aria-label</code> attribute with a text description to the interactive element itself will be sufficient to provide an accessible alternative name for the element, as well as providing the mouse tooltip.
```
adildostmohamed: @talbs When using the aria-label approach this does not create a tooltip that you would get using the title attribute and so it might be better to encourage to create their own tooltips that reveal on focus of the element as well as on focus as well which is a limitation of the title attribute I believe but would benefit keyboard only users to be able to see a text based description using a custom tooltip

src/_includes/accessibility/accessibility-manual.html
```diff
@@ -93,19 +93,18 @@ <h4 class="margin-top-xl text-muted">If an icon represents an interactive elemen
...
   <i class="fa fa-bars" aria-hidden="true"></i>
 </a>
 {% endhighlight %}
     <small class="text-muted">an icon being used as a link to a navigation menu</small>
 
 {% highlight html %}
-<a class="btn btn-danger" href="path/to/settings" title="Delete">
-  <i class="fa fa-trash-o" aria-hidden="true"></i>
+<a class="btn btn-danger" href="path/to/settings" aria-label="Delete">
+  <i class="fa fa-trash-o" aria-hidden="true" title="Delete"></i>
```
adildostmohamed: as per the comment in the thread from @karlgroves we should be careful about encouraging the use of the title attribute in case people don't understand its limitations (perhaps encouraging them to use a custom tooltip for this purpose) for accessibility reasons but strictly speaking the code you're proposing isn't incorrect

talbs: @adildostmohamed, thanks for the feedback. I wanted to leave one example present to show that a native tooltip (via `title` attr) can be added to this recommended syntax. I'll make that difference clear as well as leave a note about custom tooltips.

src/_includes/accessibility/accessibility-manual.html
talbs: Thanks for the catch. I'll correct that statement.

talbs:
@adildostmohamed, I addressed your good feedback. Thanks again for taking the time. Is there anything else you or @karlgroves think this needs before we get this good adjustment merged?

css/font-awesome.css
```diff
@@ -61,7 +61,7 @@
 }
 .fa-border {
   padding: .2em .25em .15em;
-  border: solid 0.08em #eeeeee;
+  border: solid 0.08em #eee;
```
tagliala: it looks like an unwanted change, maybe a different version of the less compiler?

There is no difference at all, but changes to the main .css could be removed from this PR to keep it clean

talbs: Sure, I included it in a separate commit (https://github.com/FortAwesome/Font-Awesome/pull/9586/commits/1dbb74d666d1906d77abe86b3c6ce9e9dd5071cf) - I'll remove that and open a new PR to abbreviate those values.

talbs:
@tagliala, I removed that extra CSS code change. This seem good from your perspective? Worth waiting for official reviews from the other folks mentioned here or just merge things in?

tagliala:
@talbs it is ok for me and doc seems clear to me

@karlgroves @adildostmohamed could you please take a final look?

adildostmohamed:
@talbs @tagliala , looks good to me following the changes you made to the content. Is there somewhere I need to click to approve the PR or something?

talbs:
Thanks for the review, @adildostmohamed. @tagliala given that we've got 2 out of 3 thumbs and to respect @karlgroves's time, I'm going to merge this work in.

All, please shout if you need anything else. Thanks for continuing to make improvements on the a11y side. :)

PR #9638, opened 2016-07-25
Updating Campaign URL Params with Latest Version Number

talbs:


talbs:
@davegandy, updated all URL params to match our current version. The only updates that couldn't be automated with the `minor_version` variable from `config.yml` was the JS-based banner ads.

PR #9732, opened 2016-08-10
Mixins to select icons to include

arondn2:
I think that is good idea have a way to include only some icons, for example:

``` scss
$fa-included-icons: (
  glass,
  camera,
  print,
);
@import 'font-awesome/scss/font-awesome';
```


ketan:
+1 — this will also help fix #6020

robmadole:

– FA team

arondn2:
Great @robmadole. It is not problem, it was just a suggestion.

Issue #9793, opened 2016-08-23
Broken link success page


ilpatrol:
"Try Again" link doesn´t work in success page when submit email for CDN

tagliala:
From https://cdn.fontawesome.com/register/success I'm able to click on "Try again" (home page).

I think you should insert your email again. It doesn't seem a bug, let's ask @talbs if this behavior is intended

Devinsuit:
"Try again" button doesn't work  in http://fontawesome.io/cdn/success/

tagliala:
@Devinsuit thanks, got it.

@talbs there are broken links at:
* https://github.com/FortAwesome/Font-Awesome/blob/master/src/cdn/error.html#L38
* https://github.com/FortAwesome/Font-Awesome/blob/master/src/cdn/success.html#L25

Don't know where you want to redirect users, maybe you could hardcode https://cdn.fontawesome.com/

talbs:
Thanks for the heads up, folks. I'm on it and will post a fix shortly.

PR #9808, opened 2016-08-24
Fixing Broken CDN "Try Again" Links

talbs:
See https://github.com/FortAwesome/Font-Awesome/issues/9793.


talbs:
FYI, @tagliala. I'm merging this and kick off a quick deploy with the fix.

PR #10061, opened 2016-10-13
Use fragment links instead of dead links

MacroChip:
Instead of using dead links to demonstrate buttons, I changed them to be fragment links. This way, users can also press the buttons with no consequences. Both sections now validate.

### DIFF
--- a/src/icons.yml
+++ b/src/icons.yml
@@ -1840,6 +1840,7 @@ icons:
       - settings
       - fix
       - update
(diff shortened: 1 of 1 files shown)


tagliala:
Could you please rebase on the current master (which already includes #10062) and squash commits?

MacroChip:
Yes. I can squash them to 1 commit and resubmit the PR.

PR #10062, opened 2016-10-13
Replaced UTF codes with html entities.

MacroChip:
My build was dying with the output:
Liquid Exception: invalid byte sequence in US-ASCII in _includes/examples/bordered-pulled.html, included in examples.html

I used the changes in these commits to move past the issue to continue development. If this PR is not accepted, do developers need to change some setting to use UTF-8 instead of US-ASCII?

I have only tested this on Chrome 53.0.2785.143 (64-bit) and Firefox 49.0 for linux mint.
I see that http://www.w3schools.com/charsets/ref_utf_punctuation.asp warns that some browsers may not support some entities. I do not know if these changes are unsupported entities.


tagliala:
Merged, thanks!

PR #7253, opened 2015-08-18
feat(bower): add css and fonts for use with bower tools

wzr1337:
Using font-awesome with bower tool like https://github.com/ck86/main-bower-files needs some more details regarding actual dist files

PR #10212, opened 2016-11-12
Adding Fonts and CSS to be tracked by Main Bower Files

cirocfc:
So, as the title suggests, I added the font files and the css file to the main property in the bower.json file.

I did it because I was struggling to perform a simple task of working the font files and copying it to my dist folder using a plugin called [main-bower-files](https://www.npmjs.com/package/main-bower-files).

When searching about this problem, I found an explanation on how the plugin works and how it maps the dependencies:

image

So, in order to make it less painful and more automatic the process of mapping dependencies, I suggest adding these changes to main property, so the plugin will be able to map such files as well as .scss and .less files that are already been mapped.

Thanks 😄

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -7,8 +7,15 @@
   "devDependencies": {},
   "license": ["OFL-1.1", "MIT", "CC-BY-3.0"],
   "main": [
+    "css/font-awasome.css",
     "less/font-awesome.less",
-    "scss/font-awesome.scss"
+    "scss/font-awesome.scss",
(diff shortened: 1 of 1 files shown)


wolfy1339:
This is invalid as per the bower spec. Please search more in the future because they're alot of duplicate issues and pr that want to do the same thing. They explain the reasoning and the proper lunks to bower spec

cirocfc:
Related: https://github.com/FortAwesome/Font-Awesome/pull/7253

PR #10307, opened 2016-12-01
Update pseudo elements syntax

### LINKED ISSUE #9582: Use double-colon CSS3 syntax for pseudo elements
standard font-awesome.css has following declarations:

`.fa-camera-retro:before {  content: "\f083"; }`

which technically incorrect as `before`  is [pseudo-element but not pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements). Therefore those declarations all should look like as

`.fa-camera-retro::before {  content: "\f083"; }`

While `:before` form is supported as an exception in most browsers I think for standard conformity that CSS should be changed.

mrmohtas:
Update `:before` to `::before` to be inline with the correct usage of pseudo element syntax

Resolves #9582


tagliala:
Sorry, cannot merge this change, this is a breaking change for IE8, which apparently will be also supported in FA 5.0 :/

Also, please take a look here: https://github.com/FortAwesome/Font-Awesome/blob/master/CONTRIBUTING.md#pull-requests

> Any changes to the styles must be made to the .less and .scss files in the src directory

Issue #10245, opened 2016-11-18
Proper Unicode Font Variant


Crissov:
As a variant of #222, could you please release a version of Font Awesome (5) that *only* includes characters present in Unicode with their proper code-points, no PUA? This would, of course, mean that a lot of FA glyphs would not be available at all in that variant.

If that is outside the scope of this project, is there, by any chance, already a fork for something like that?

sensibleworld:
Doing a little management in Github and (batch) closing out potentially stale issues. If the above still needs attention, please let us know and we’ll re-open!

Issue #222, opened 2012-06-04
Support public Unicode codepoints where possible


eevee:
It would be fantastic if I could use codepoints for which appropriate icons already exist, e.g., "★" for the solid star.

A hypothetical screenreader could even then announce the name of the glyph.

StanleyMlha:
+1 : map to Unicode possible code-points

karlhorky:
What's the status of this? Is this still planned for version 5? I took a look and didn't find a pull request for this yet - I'm assuming there isn't one yet due to difficulty of a clean solution.

Is [this](https://github.com/FortAwesome/Font-Awesome/milestones/5.0.0) is the only tracking page for version 5?

joewreschnig:
There is absolutely a clean solution: Put the characters at both the public and private codepoints.

There is no pull request because the fonts are checked in as binary files and it is impossible to submit patches for them.

dwt:
I was directed here to re-report my problem, so here goes:

My main point is that we have found that windows has a 'feature' that allows admins to disable all webfont downloads for all their users. This leads to apps that use font awesome to suddenly loose all icons from the application (in all versions of IE!), usually rendering the app completely unusable. See https://github.com/FortAwesome/Font-Awesome/issues/10046 for the original bug report.

I also think that together with support for content-editable and screen readers reading out the correct characters, it makes _a lot of sense_ to switch to semantic code points, and I'd like to add my voice here.

Crissov:
Will this finally be fixed in Font Awesome 5 – could it be added as a goal to [Kickstarter](https://www.kickstarter.com/projects/232193852/font-awesome-5/)?

Crissov:
Has it been fixed in v5.0.0, the milestone this issue is tagged with? I only see PUA codes in the documentation.

tagliala:
This should be finally available in FA 6 beta

sensibleworld:
Yup! We've been working hard to map our existing icons to standard unicode points. In `6.0.0-beta2` they have been assigned as aliases. We still need to reveal these code points to the end-user, but they exist. (And we still have a ways to go for full parity.)

anarchodin:
@sensibleworld Is that mapping visible to the public anywhere? I'm guessing you could get help finding assignments that are missing.

[...]

PR #10357, opened 2016-12-13
Existing Unicode code points per YAML

Crissov:
- Shows how to implement #10245 .
- Helps to address #222. Compatible with most suggestions in [this comment from 2013](https://github.com/FortAwesome/Font-Awesome/issues/222#issuecomment-16621596) by @lpar.


tagliala:
@Crissov I want to thank you for your big effort here and in all other issues where you are providing the standard unicode point of the request. ❤️ 

This PR is quite impressive and it is a very good starting point for a brainstorming about standard unicode points

This choice is far out from my tasks, but I can definitely point out a couple of things:
1. The `.svg` file is automatically generated from the `.otf` file via font squirrel. So, standard unicode points should be added to the `.otf` (and we are not accepting PR on that file, mostly because of its binary nature);
2. This is _probably_ a breaking change. I know that we could update our build scripts to reflect this change, but we don't know what other users may be doing with that `.yml`. So, the best thing we can do is to consider another way of adding standard unicode points to the .yml file, leaving `unicode:` as it is.  I can't actually think of a better name, maybe just a simple array `std_unicode`.
  E.g. 
  ```yml
  - name:       Check
    id:         check
    unicode:    f00c
    std_unicode:
      - 2713 # ✓
      - 2714 # ✔
  ```

let me know your thoughts

Crissov:
Since I couldn’t patch the OTF, the YML was the best place I could find. The pull request may have been premature, as it’s a work in progress, but I needed some feedback indeed, e.g. whether the `unicode:` / `pua:` split was acceptable. I’ll change it in the next couple of days.

Some FA glyphs could apply to several codepoints and some Unicode codepoints have several glyphs that could map to them – that needs to be straightened out. In a proper font, `outline`, `square` and `circle` variants should probably have glyph names where these are suffixes with a leading dot `.` and they should be accessible with `ss##` or `cv##` features. There would also have to be some ligation code for overlays, perhaps relying on ZWJ (U+200D) for automatic, mandatory ligatures.

Manually editing the YAML and SVG files, I noticed (and corrected if possible) several minor glitches, e.g. some stray empty glyphs. Somewhere in the process of conversion to SVG, many glyph names seem to get messed up.

robmadole:

Over the past year we've been hard at work on Font Awesome 5. It's now released and this PR no longer applies _directly_.

@Crissov this looked like a massive amount of work. And I don't want to completely throw this away.

However one of the things that we need to evaluate before we consider a change like this is the use case for aligning these unicode values with the official ones. We really struggled to come up with a good reason to do this. For many icons there just were not obvious choices. We'll also have icons in Font Awesome that do not have an official unicode value. And the backward-incompatible nature of this is really scary.

So if you are interested in exploring this with us (and why you thought this was necessary) we'll put this idea back on the table.

– FA team

Crissov:
The ancient but still valid issue #222 is about *also* assigning standard Unicode code points to FA glyphs *where possible*, so authors can choose to use standard characters which will fallback to system fonts if FA is not available for some reason. Some of them are colorful emojis, but it is the author’s job to decide whether they can live with 1. that fallback and 2. the fact that some platforms do not support custom fonts for emoji characters at all.

My more recent issue #10245 is about a restricted version of FA that has removed all PUA assignments as well as all glyphs without a standard code point. The use case is similar. Authors would know that not everything FA offers would be available on this font file.

robmadole:
@Crissov thanks for being patient with us and working through this.

_We intended to add "proper" unicode in 5.0 but we couldn't find a clean solution_.

The issue is a little deeper than just adding the Unicode code points and then re-generating the web fonts. You said "it changes absolutely nothing for people who are (and keep on) using the PUA codes". You are absolutely correct. However, we have to think about this from not just a technical perspective. This feature has to be supported and documented.

unicode-heart-example

I'll use this [screen grab](https://unicode-table.com/en/sets/hearts-symbols/) of the "heart" symbols to illustrate some of the issues that we have ran into that led us to abandon this effort with version 5.

1. There are multiple "solid" hearts, which one is canonical? Is the canonical version chosen by appearance? By glyph name? Do we map multiple code points? If I put the wrong unicode heart in my page and Font Awesome doesn't work is that a bug or my mistake?
1. For our outlined styles Font Awesome Regular and Font Awesome Light it does not seem appropriate for those to map to any of the "visually solid" Unicode glyphs. We'd need to map the Light heart to "White Heart Suit"? So this means that changing between styles using CSS (using font-weight) or in a desktop program is not possible. If you wanted to go from FA Solid "heart" to FA Regular "heart" the code point would need to change.
1. If we chose "Black Heart Suit" (which would make sense if we map to "White Heart Suit" for solid) as the canonical version because it visually matches the icon in Font Awesome there is an expectation that we would have the other symbols from the group as well (club, diamond, spade). If would also place a constraint on these icons (they need to look like the typical symbols on a deck of cards) that may not be desirable.
1. For the icons that do not have an associated Unicode code point, this creates a documentation and problem for anyone that has decided to use this particular technique to show icons. If they are relying on fallback fonts to display standard Unicode symbols and the icon is non-decorative we will not have a solution for this.

> Some of them are colorful emojis, but it is the author’s job to decide whether they can live with 1. that fallback and 2. the fact that some platforms do not support custom fonts for emoji characters at all.

This level of user intervention/responsibility is atypical for Font Awesome. We want to continue to make things as easy as possible. This doesn't strike me as within the same spirit of what Font Awesome is.

> My more recent issue #10245 is about a restricted version of FA that has removed all PUA assignments as well as all glyphs without a standard code point. The use case is similar. Authors would know that not everything FA offers would be available on this font file.

This would be a heavy burden for authors. And what recourse do they then take if they need an icon that is not included in the Unicode code points?

> I feel like I have to emphasize something that has been pointed out repeatedly in #222 before: it changes absolutely nothing for people who are (and keep on) using the PUA codes.

I disagree with this. It adds layers of [complexity and decision points](http://uxmyths.com/post/712569752/myth-more-choices-and-features-result-in-higher-satisfac). The fewer options that we have for using Font Awesome the easier and more enjoyable it will be to use.

I want to emphasize that this topic is not something that we take lightly. We've put a fair amount of thought into it and our team was not able to solve it (I would say we've failed to solve it/understand it).

So in that vein, we'll need to table this when we can come back to it and give it the proper attention. We underestimated it from the beginning.

Project management note:

* #222
* #10245
* #10357 

Will be used as reference when we begin working on this again.

Crissov:
The point of these issues is that some people want to use Font Awesome as an actual font, not as an icon set packaged as a font file.

PR #10374, opened 2016-12-15
Update license to include CDN assets

robmadole:


robmadole:
@davegandy we didn't have any license information for CDN. What do you think about this?

Issue #10395, opened 2016-12-18
Cheatsheet print problem


aromot:
When trying to print in PDF the cheatsheet at the following address :

http://fontawesome.io/cheatsheet/

I only get the 3 first pages instead of the full list of icons (tested in Chrome 55 & Opera 42).

tagliala:
I can confirm. I will try to fix this

tagliala:
@armot I've prepared a fix at #10396. Meanwhile, I've attached there a proper PDF with all icons

tagliala:
Thanks. The fix was committed to master, it will be live when the docs will be rebuild

PR #10396, opened 2016-12-18
Fix cheatsheet print layout

tagliala:
This PR changes the print layout, showing three icons per row instead of
one. In this way we can solve an issue with the Bootstrap grid, which
for some reason limits the output to three pages when printing the
cheatsheet.

Fix: #10395

Before: [Font Awesome Cheatsheet.pdf](https://github.com/FortAwesome/Font-Awesome/files/659571/Font.Awesome.Cheatsheet.pdf)

After: [Font Awesome Cheatsheet (2).pdf](https://github.com/FortAwesome/Font-Awesome/files/659572/Font.Awesome.Cheatsheet.2.pdf)

### DIFF
--- a/src/cheatsheet.html
+++ b/src/cheatsheet.html
@@ -23,7 +23,7 @@ relative_path: ../
     {% assign sorted_icons = icons | expand_aliases | sort_by:'class' %}
 
     {% for icon in sorted_icons %}
-    <div class="col-md-4 col-sm-6 col-lg-3">
+    <div class="col-md-4 col-sm-6 col-lg-3 col-print-4">
       {% if icon.created >= site.fontawesome.major_version %}<small class="text-muted pull-right">{{ icon.created }}</small>{% endif %}
       <i class="fa fa-fw" aria-hidden="true" title="Copy to use {{ icon.class }}">&#x{{ icon.unicode }}</i>
       fa-{{ icon.class }}
--- a/src/cheatsheet.html
+++ b/src/cheatsheet.html
@@ -25,7 +25,7 @@ relative_path: ../
     {% for icon in sorted_icons %}
     <div class="col-md-4 col-sm-6 col-lg-3 col-print-4">
       {% if icon.created >= site.fontawesome.major_version %}<small class="text-muted pull-right">{{ icon.created }}</small>{% endif %}
-      <i class="fa fa-fw" aria-hidden="true" title="Copy to use {{ icon.class }}">&#x{{ icon.unicode }}</i>
+      <i class="fa fa-fw" aria-hidden="true" title="Copy to use {{ icon.class }}">&#x{{ icon.unicode }};</i>
       fa-{{ icon.class }}
       {% if icon.alias_of %} <span class="text-muted">(alias)</span>{% endif %}
       <span class="text-muted">[&amp;#x{{ icon.unicode }};]</span>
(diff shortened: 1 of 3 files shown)


src/cheatsheet.html
```diff
@@ -23,9 +23,9 @@ <h2 class="page-header">Every Font Awesome {{ site.fontawesome.version }} Icon,
     {% assign sorted_icons = icons | expand_aliases | sort_by:'class' %}
 
     {% for icon in sorted_icons %}
-    <div class="col-md-4 col-sm-6 col-lg-3">
+    <div class="cheatsheet-icon">
       {% if icon.created >= site.fontawesome.major_version %}<small class="text-muted pull-right">{{ icon.created }}</small>{% endif %}
-      <i class="fa fa-fw" aria-hidden="true" title="Copy to use {{ icon.class }}">&#x{{ icon.unicode }}</i>
+      <i class="fa fa-fw" aria-hidden="true" title="Copy to use {{ icon.class }}">&#x{{ icon.unicode }};</i>
```
robmadole: @tagliala what's the semicolon for?

robmadole:
This looks good to me. Nice fix. I'll defer to @talbs on the bootstrap usage. I wonder if they have a breakpoint class that will accomplish what you did with additional CSS...

src/cheatsheet.html
talbs: @tagliala I don't think the semicolon is needed here. Its needed a bit below (line  31), but the current site renders the icons just fine without it.

src/cheatsheet.html
```diff
@@ -23,9 +23,9 @@ <h2 class="page-header">Every Font Awesome {{ site.fontawesome.version }} Icon,
     {% assign sorted_icons = icons | expand_aliases | sort_by:'class' %}
 
     {% for icon in sorted_icons %}
-    <div class="col-md-4 col-sm-6 col-lg-3">
```
talbs: Is there any way to leave these classes in for screen rendering and as importantly transparency? Can we target and override the print styling with the print media-query-based rules only?

src/assets/less/site/cheatsheet.less
```diff
@@ -0,0 +1,9 @@
+.cheatsheet-icon {
+  .make-sm-column(6);
```
talbs: As mentioned in `cheatsheet.html`, I'd love to have this new less partial only have to contain the print-specific styling for the cheatsheet. Can you try the following?

1) removing the non-print rules here and re-instating the `col-md-4 col-sm-6 col-lg-3` classes in the `cheatsheet.html` view.
2) keep the print-specific rule in this file and confirm that it is obeyed when the cheatsheet is printed.

src/cheatsheet.html
tagliala: It is explained in a separate commit. The W3C validator complains:

image

src/cheatsheet.html
tagliala: @talbs please take another look

I could add a full grid for the print layout, but at the moment we only need the col-4 helper

talbs: That makes sense to me. This seems the right scope for what you're doing here.

tagliala:
@talbs @robmadole thanks for the review!

I'm not confident in rebuilding the docs and pushing them to the `gh_pages` branch, but since I've attached a proper pdf in the opening post, we can wait for the next docs rebuild to see this fix live

PR #10406, opened 2016-12-21
Allow custom naming of font family

james-johnston-thumbtack:
This allows multiple variations/versions of Font Awesome to be used,
similar to the namespacing currently allowed by the $fa-css-prefix
variable.

(And also allows us to more easily use a third-party library that makes
hard-coded non-default assumptions about the font face name.)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -29,6 +29,7 @@ fontawesome:
   twitter:        fontawesome
   tagline:        The iconic font and CSS toolkit
   css_prefix:     fa
+  font_family:    FontAwesome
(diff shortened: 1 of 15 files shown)


tagliala:
IIRC, this was discussed before and Dave choosed to not implement this feature.

At the moment I'm unable to find the original request, maybe it was something discussed internally.

BTW leaving this open

tagliala:
Yep, found it: #6113

I'm going to close this one but I want to thank you for this PR before!

PR #10455, opened 2017-01-02
Added animation keyframes

landsman:
I have pseudo element and need rotation (spinner):

**before commit:**

```
:after {
      content: $fa-var-spinner;
      font-family: "FontAwesome";
      @include fa-icon-rotate(90deg, 1);
      animation: fa-spin 2s infinite linear;
}
```

**after my changes:**

```
:after {
      content: $fa-var-spinner;
      font-family: "FontAwesome";
      @include fa-icon-rotate(90deg, 1);
}
```

### DIFF
--- a/scss/_mixins.scss
+++ b/scss/_mixins.scss
@@ -12,6 +12,7 @@
 }
 
 @mixin fa-icon-rotate($degrees, $rotation) {
+  animation: fa-spin 2s infinite linear;
   -ms-filter: "progid:DXImageTransform.Microsoft.BasicImage(rotation=#{$rotation})";
   -webkit-transform: rotate($degrees);
       -ms-transform: rotate($degrees);


tagliala:
Hi,

thanks for being part of the Font Awesome community.

Unfortunately I cannot leave this PR open because:
- Files under the `/less` and `/scss` folders are automatically generated by the sources in [`/src`](https://github.com/FortAwesome/Font-Awesome/tree/master/src/assets/font-awesome)
- This PR only fixes the Scss mixin, leaving out the [Less one](https://github.com/FortAwesome/Font-Awesome/blob/master/src/assets/font-awesome/less/mixins.less#L10)

Also, I don't exactly get what this PR is trying to achieve. It seems that it is adding a spinning animation to the `.fa-rotate` mixin, and this doesn't look semantic. I mean, if I want to apply a rotation to an icon, I probably don't want that icon to spin too.

What this PR was trying to fix? Could you please provide a fiddle https://jsfiddle.net/tagliala/s3msxtbe/ showing your issue?

landsman:
I was mean that "fa-icon-rotate" should contain "fa-spin" animation.

PR #10803, opened 2017-03-17
Make icon names more human friendly.

Mayyay:
A fair few of the icon names in `icons.yml` seemed to be the same as the icon IDs.

I'd like to propose these changes as it will also make it easier to automatically generate user friendly names for use in the likes of icon pickers.

### DIFF
--- a/src/icons.yml
+++ b/src/icons.yml
@@ -110,7 +110,7 @@ icons:
     categories:
       - Web Application Icons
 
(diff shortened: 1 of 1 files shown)


tagliala:
It is ok for me. It should not affect docs, because `name` seems unused

robmadole:

– FA team

PR #11020, opened 2017-05-03
Fix Composer package "type" to "component"

glensc:
This will allow easy deployment into a public directory with [component-installer](https://github.com/RobLoach/component-installer) or other various methods.

related links:
- https://github.com/RobLoach/component-installer/issues/98
- https://github.com/RobLoach/component-installer/issues/75
- https://github.com/enyo/dropzone/pull/857

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,7 @@
   "description": "The iconic font and CSS framework",
   "keywords": ["font", "awesome", "fontawesome", "icon", "font", "bootstrap"],
   "homepage": "http://fontawesome.io/",
+  "type": "component",
   "authors": [
     {
       "name": "Dave Gandy",


tagliala:
Hi!

Thanks for being part of the Font Awesome Community and for this PR.

I'm not into composer and I'm trying to understand.

I'm wondering why if it is a feature or a bugfix. I'm not seeing this in the [composer.json](https://github.com/twbs/bootstrap/blob/v4-dev/composer.json) of bootstrap.

Also, it seems not being a standard: https://getcomposer.org/doc/04-schema.md#type

Why we should use the undocumented `"type": "component"` instead of the default value `library`?

Thanks in advance

glensc:
as i understood, it's needed for [RobLoach/component-installer](https://github.com/RobLoach/component-installer) to find packages to copy components out.
the composer documentation that you referred does not say the list is limited to what's written on that page:
> Package types are used for custom installation logic. If you have a package that needs some special logic, you can define a custom type. This could be a symfony-bundle, a wordpress-plugin or a typo3-cms-extension.

so here's package that needs that "special logic".

also "library" in composer context is intended for projects that store "php library code", `Font-Awesome` is not that :)

@RobLoach can probably comment more

RobLoach:
Seems good to me. Type is only used by Composer Installers that act on certain types, otherwise they just default to "library". If "component" is not defined by any composer installers, it's assumed to be a "library".

tagliala:
Thanks for the clarification.

I cannot merge this, I'm going to ask for a review from Rob or Dave before

robmadole:
I don't have any experience with this but it seems like a benign change. :+1:

PR #11027, opened 2017-05-06
Add filter "email" to "at"

doersino:
Makes "at" show up when searching for "email" on [fontawesome.io/icons](http://fontawesome.io/icons/).


tagliala:
Thanks!

PR #11550, opened 2017-09-23
The .fa- are now generated with an array with the PUA codes and icon names

pjotrvdh:
Easy function to get the PUA codes for an icon: fa-var(pencil) or with map-get($fa-var, pencil)
That way you don't have to manually add the :before { content: $fa-var-* } for each icon during development.


robmadole:
This is good stuff. But unfortunately, I think it would not be backward-compatible. If I'm mistaken let me know.

robmadole:

– FA team

PR #11808, opened 2017-11-28
Added "main" property

franciscop:
Right now to import Font-Awesome into a React project it has to be done like this (after `npm install font-awesome`):

```js
import React from 'react';
import ReactDOM from 'react-dom';
import 'picnic';
import 'font-awesome/css/font-awesome.min.css';

// ...
```

However, as you can see from the line before, I can import my own library Picnic CSS with just `import 'picnic'`. This is because [inside Picnic CSS' package.json](https://github.com/franciscop/picnic/blob/master/package.json#L11) I have set this property:

```js
 "main": "picnic.min.css",
```

So I am proposing to add the same property for Font-Awesome to be able to import Font-Awesome with just this code:

```js
import React from 'react';
import ReactDOM from 'react-dom';
import 'picnic';
import 'font-awesome';

// ...
```

This will make it really intuitive for React devs to install and use Font-Awesome without having to dig into the code/repository to find out where the actual package is located.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "font-awesome",
   "description": "The iconic font and CSS framework",
   "version": "4.7.0",
+  "main": "css/font-awesome.min.css",
   "style": "css/font-awesome.css",
   "keywords": ["font", "awesome", "fontawesome", "icon", "bootstrap"],
   "homepage": "http://fontawesome.io/",


tagliala:
Hi!

Thanks for submitting this

Duplicate of #8368

I'm discussing this also in the Pro repo for 5.0 but I'm not authorized to make this decision and merge this PR, sorry

franciscop:
@tagliala thanks for reviewing this, sorry I didn't see that issue. As I understand it, wouldn't this be the actual implementation of #8368 and not just a duplicate? Discussion of yes/no over there while keeping this about the implementation details.

tagliala:
@franciscop sorry, you are right.

This should be a duplicate of #9430, but as you can see #9430 is trying to put the `.scss` file as the main entry point

I can confirm this issue has been submitted to the Font Awesome Pro repo.

Please use #8368 for discussion too, +1s comments should not be allowed anymore (there are reactions for upvoting)

robmadole:
Font Awesome 5 changes a lot of this landscape. We'll most likely need to release separate packages specifically for using web fonts (as we will have a JavaScript SVG version as well). We'll have to revisit this soon after the full release of 5.

PR #11824, opened 2017-12-01
Update icons.yml

grvrg:
consolidate Viacoin formatting to be in line with formatting of Bitcoin and other currencies

### DIFF
--- a/src/icons.yml
+++ b/src/icons.yml
@@ -5261,8 +5261,6 @@ icons:
     unicode:    f237
     created:    4.3
     url:        viacoin.org
-    aliases:
-      - via
     categories:
       - Currency Icons
       - Brand Icons
--- a/src/icons.yml
+++ b/src/icons.yml
@@ -5256,12 +5256,15 @@ icons:
     categories:
       - Web Application Icons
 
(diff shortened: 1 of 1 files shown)


src/icons.yml
```diff
@@ -5256,12 +5256,15 @@ icons:
     categories:
       - Web Application Icons
 
-  - name:       Viacoin
+  - name:       Viacoin (VIA)
     id:         viacoin
     unicode:    f237
     created:    4.3
     url:        viacoin.org
+    aliases:
```
wolfy1339: This is not necessary

grvrg: Bitcoin is listed as **Bitcoin (BTC)**, other currencies are also listed with their short symbols such as **Euro (EUR)**

wolfy1339: I wasn't talking about the whole update, just that the alias is not needed as it's the same as the id

grvrg: i see :)
so it should be probably changed as follows?
id to **via** and alias to **viacoin**

tagliala:
Thanks!

PR #12402, opened 2018-02-20
Update fa-regular.scss

danrcoull:
Names should be different for font families between fa regular, and fa solid otherwise you can only use one or the other.

I understand that:


robmadole:
Hi @danrcoull they use different weights. If you have a test case where you can show that this is broken feel free to file an issue.

PR #12760, opened 2018-04-02
Add 5.1 instructions

robmadole:


UPGRADING.md
```diff
@@ -6,6 +6,114 @@ This guide is useful to figure out what you need to do between breaking changes.
...
+
+### Improved support for tree shaking
+
+Tree shaking is now functional by default and no additional configuration is required to make it work.
+
+The `shakable.es.js` module has been removed and is no longer needed.
+
+If you've previously configured tree shaking by modifying your webpack or rollup you can safely remove these.
+
+**We recommend that you check your bundle size after upgrading an ensure that file sizes are as you would expect.**
```
mlwilkerson: I've thought this every time I've read this statement in one of the docs, and maybe I've already mentioned it before: it feels like there should be some more definitive guidance we could provide to help people verify that the bundles are being tree shaken. Depending on how sophisticated the developer is, it may be hard are confusing to know whether tree shaking worked based only on checking bundle size. When I do it, I look at the code make sure  I see only the icons I know I expect in there. But I know what JS code to look for.

Kinda makes me want to provide a little utility that knows how to look at a bundle and report which icons it has in it. Or, in the absence of such a tool, might we need to say more specifically what to look for to verify that the tree has been shaken?

robmadole: Yeah I know this is a cop-out almost. I just can't think of a good way to explain this without it getting crazy deep and complex. I'm super hoping that we can just put this to bed and the last glowing embers of this fire will extinguish. If it doesn't let's deal with it then.

mlwilkerson: k

PR #13007, opened 2018-05-02
Fix library.add() typos

mlwilkerson:


tagliala:
@mlwilkerson please also take a look at:

FortAwesome/Font-Awesome#13009
FortAwesome/Font-Awesome#13010

PR #13055, opened 2018-05-07
Fix typo in README

nikai3d:
I understand that:


tagliala:
Thanks! @robmadole is this safe to merge or it should be fixed in the build system?

robmadole:
@tagliala this is safe to merge!

PR #13155, opened 2018-05-15
In IE11 double focus bug (accessibility related)

gtibrett:
In IE11 if you tab through links (or any focusable element) containing SVG elements it will have two tab stops.

Fix is to add focusable="false" to the SVG element.
This works around an IE11 bug that defaults SVG elements to focusable="true".
The double focus can also cause some screen readers to double-read.

Example Page: https://labs.levelaccess.com/index.php/Remove_SVG_from_Tab_Order
Also See: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8090208/

I understand that:

### DIFF
--- a/advanced-options/use-with-node-js/fontawesome/index.es.js
+++ b/advanced-options/use-with-node-js/fontawesome/index.es.js
@@ -1081,6 +1081,7 @@ var attributesParser = function (node) {
       extraAttributes['aria-labelledby'] = config$1.replacementClass + '-title-' + nextUniqueId();
     } else {
       extraAttributes['aria-hidden'] = 'true';
+      extraAttributes['focusable'] = 'false';
     }
   }
 
@@ -1260,6 +1261,7 @@ function generateLayersText(node, nodeMeta) {
 
   if (config$1.autoA11y && !title) {
     extra.attributes['aria-hidden'] = 'true';
+    extra.attributes['focusable'] = 'false';
   }
 
   return [node, makeLayersTextAbstract({
@@ -1659,6 +1661,7 @@ var icon = resolveIcons(function (iconDefinition) {
         attributes['aria-labelledby'] = config$1.replacementClass + '-title-' + nextUniqueId();
       } else {
         attributes['aria-hidden'] = 'true';
+        attributes['focusable'] = 'false';
       }
     }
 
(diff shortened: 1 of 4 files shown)


tagliala:
@gtibrett thanks!

tagliala:
is this also affecting Edge < 14.14393 ?

gtibrett:
We tested in Edge 17.17134.  The IE bug is not present there.

The linked Microsoft defect states that it is fixed in 14.14393

> Fixed in EdgeHTML 14.14393

robmadole:
Ok, we'll have to be careful with this one. It's a simple code change but could have ripple effects. I'll get this one on the board. Thanks @gtibrett !

bliles:
Any updates on this? It's still an issue currently that affects our users. I would open a new PR since there are conflicts here, but given that this repo is used only for reference/publishing there doesn't seem to be much value to that since the fix is simple to implement.

bliles:
Also, just a detail since we might have the tendency to dismiss IE issues, most JAWS users are likely still using IE: http://www.hollier.info/browserpairing/

robmadole:
Alright folks, sorry for taking so long. We are catching up on older issues as we can.

So I hesitate to add `focusable` because it doesn't seem to be the future direction of SVG. The newer spec uses the same type of focusing mechanism as normal HTML elements.

Right now you can add an icon and specify this yourself using the extra attribute pass through. Like this:

```
<i class="fas fa-beer" focusable="false"></I>
```

Would this work? If this is a valid workaround I'd be happy to add to [our accessibility docs](https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility).

stevefaulkner:
FWIW adding `focusable="false"` to non interactive SVG elements is standard practice to fix the issue that occurs in IE. Note this issue also effects no screen reader keyboard users in IE. The inclusion of  `focusable="false"`  has no known negative effects.

scottaohara:
@stevefaulkner beat me to it :)

Another thing to consider, per the previously [linked accessibility docs](https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility#web-fonts-decorative)

```html
<i class="fas fa-camera-retro" aria-hidden></i>
```

snippets like that should probably be updated to specifically call out `aria-hidden="true"`  Without a set value, `aria-hidden` defaults to undefined and screen readers can still parse the content.

tagliala:
@scottaohara thanks for reporting this. Could you please open a new issue?

scottaohara:
@tagliala created #14506

robmadole:
@gtibrett @bliles @jimjenkins5 @stevefaulkner ok, you all have convinced me. My only concern with making it the default is the potential to create a problem for someone else. You all are telling me there are no downsides so I'll take that bet with ya. Let's :shipit:.

This should be implemented as part of 5.7.0.

bliles:
Confirmed fixed with 5.7.0.

tagliala:
Hi folks, 👋 👋 

Any thoughts about this ➡️ #14595  ?

PR #13318, opened 2018-06-13
To make it easier, I added requesting icons right to the README.md

RDIL:
Yeah that's it.  If you don't like these changes its fine because I don't care.  Just suggesting this.


Killerbear:
There is already a link for requesting new icons just one line above your added code.
I do not think that needs to be mentioned twice.
I´m not a member of the fontawesome team. But i don´t think this change will not be merged.

tagliala:
Thanks.

Maybe the Contributing section could be more explicit

PR #13415, opened 2018-06-25
Bad references for LESS shim

travisdahl:
Was getting errors trying to import the shims.  fixing these references seemed to solve it.

I understand that:


tagliala:
Thanks.

@robmadole could you please check this?

Seems legit, since https://github.com/FortAwesome/Font-Awesome/blob/master/web-fonts-with-css/less/fontawesome.less is importing partials with underscore and less extension

travisdahl:
@tagliala that was the same assessment I made, i thought maybe it was a format i didnt know about but then I looked at the other file and it seemed like it was wrong. And it did in fact fix the errors

robmadole:
Yep, good catch @travisdahl . We'll get this fixed.

PR #13840, opened 2018-08-29
replaced null by empty array

scipper:
replaced null by empty array in Folder search terms, to prevent isArray check when iterating icons

I understand that:

- [x ] I'm submitting this PR for reference only. It shows an example of what I'd like to see changed but

### DIFF
--- a/advanced-options/metadata/icons.json
+++ b/advanced-options/metadata/icons.json
@@ -14325,7 +14325,7 @@
     ],
     "ligatures": [],
     "search": {
-      "terms": null
+      "terms": []
     },
     "styles": [
       "solid",


tagliala:
Thanks!

talbs:
Thanks for the catch, @scipper. I think we need to make this change back within our icons' build system. I'll get on that now with @robmadole's help. Thanks for the help in squashing this bug. 😄

talbs:
We've got this fixed back in the build system (i.e. the factory that boxes up all of our icons). This will be out in our next release. Closing this PR with that in mind. Thanks again for the help and awesome effort, @scipper. 🙌

scipper:
I'm glad, that I could help 😊

PR #14306, opened 2018-11-22
fa-beat fa-hover-beat animations

moldypenguins:
I understand that:

adds heartbeat animation (webfonts with css only)


moldypenguins:
Similar to #13940

moldypenguins:
EXAMPLE: Heart will beat when you hover the link
<a href="#" class="fa-parent"><i class="fas fa-heartbeat fa-hover-beat"></i> Heartbeat</a>

robmadole:
Thanks for the PR but we think using something like animate.css is the best choice for animation. That way the FA lib doesn't have to get larger.

PR #14307, opened 2018-11-22
LICENSE.txt typo fix, OLF->OFL

vlna:
I understand that:


tagliala:
@vlna thanks!

There is also a typo on the docs at https://fontawesome.com/license/free

I'm assigning both @robmadole and @talbs because I don't know if this has to be fixed in the build system

talbs:
Thanks for catching this, @vlna. I've fixed the typo in our build-system that generates this LICENSE.txt file with each release. This will be fixed with our next release shortly. 

@tagliala, thanks for the heads up about the same typo on https://fontawesome.com/license/free. I've got a fix pushed for that as well and it should go out with our next site update.

Issue #14380, opened 2018-12-11
Error when deploy in server (user new version 5.6.0). The old version still works (5.5.0-2)


yugienter:
![image](https://user-images.githubusercontent.com/7371616/49785837-63e78e80-fd54-11e8-921c-d5dec38cc743.png)

tagliala:
Hi!

Thanks for being part of the Font Awesome Community.

@robmadole could you please take a look at this? Could this be related to FortAwesome/Font-Awesome-Pro#1286 ?

I personally do not have issues with deploys on my demo application:
https://github.com/diowa/ruby2-rails5-bootstrap-heroku/blob/master/package.json
https://github.com/diowa/ruby2-rails5-bootstrap-heroku/blob/master/yarn.lock

mllustosa:
Hello guys and girls,

I'm also having a similar problem when trying to deploy the newest version.

![image](https://user-images.githubusercontent.com/3597619/49804035-d9059480-fd51-11e8-9875-84186c995d75.png)

robmadole:
@yugienter I think this is related to #14382. I have the fix ready for this and will deploy in just a moment. You can test `5.6.1` and see if it resolves it. If it doesn't we'll need to get a reproducible test case from you to work on the fix.

robmadole:
This should be all fixed up now with `5.6.1`. Let us know if you still have issues.

PR #14384, opened 2018-12-11
Fix/const to var

mllustosa:
This is related to the issue #14380. The latest update used `const` variable declaration which is not compatible with UglifyJS (UglifyJS doesn't support ES6). From what I saw, those were the only places where the library is using `const` and in all other places `var` is used, so I assumed that it was by mistake.

**Of course that this is just a small draft and you guys know better what needs to be changed and where, but I thought that giving some sort of direction wouldn't hurt**

I understand that:


clark-stevenson:
Thanks for this @mllustosa 

I confirm that my webpack build is failing at the uglifyjs step because of it.  I was in the middle of writing a stack overflow about why `const` was ending up in my output which I didn't expect.

tagliala:
@robmadole has this been fixed in 5.6.1?

robmadole:
Yeah we've fixed this in `5.6.1` :+1:

tagliala:
@mllustosa thanks for your PR

mllustosa:
@tagliala Anytime. Thanks for this awesome library and for the great and quick support 😄

PR #14512, opened 2019-01-17
i2svg options should be optional

cooljeffro:
In i2svg, when specifying the 'node' param, typescript then requires the 'callback' param. Both 'node' and 'callback' should be optional.

I understand that:


tagliala:
@mlwilkerson could you please take a look here?

mlwilkerson:
Thanks @cooljeffro and @tagliala . This looks good to me and seems to bring the intended use of `i2svg()` in line with our own [API doc examples](https://fontawesome.com/how-to-use/with-the-api/methods/dom-i2svg).

I'll pass it along for possible inclusion in a future release of `fontawesome-svg-core`

js-packages/@fortawesome/fontawesome-svg-core/index.d.ts
```diff
@@ -101,7 +101,7 @@ export interface IconParams extends Params {
   mask?: IconLookup;
 }
 export interface DOM {
-  i2svg(params?: { node: Node; callback: () => void }): Promise<void>;
+  i2svg(params?: { node?: Node; callback?: () => void }): Promise<void>;
   css(): string;
   insertCss(): string;
```
siznap-jiwxeG-kocgo4, line 106: - - ```
``
```

Comment (siznap-jiwxeG-kocgo4) [inline js-packages/@fortawesome/fontawesome-svg-core/index.d.ts:106]: - - ```
``
```

BraianS:
Any update about it?

PR #14577, opened 2019-01-31
Remove vertical-align on fa-icon mixin to fix alignment

florianlacreuse:
When using `fa-icon` mixin, the displayed icon seems to not be aligned as expected because of custom `vertical-align`.

See this reduced test case: https://codepen.io/anon/pen/ErZXgv

Unless `vertical-align` is still here for a purpose (I didn't find a reason), I suggest to remove it to fix this alignment issue.

### DIFF
--- a/less/_mixins.less
+++ b/less/_mixins.less
@@ -9,7 +9,6 @@
   font-variant: normal;
   font-weight: normal;
   line-height: 1;
-  vertical-align: -.125em;
 }
 
 .fa-icon-rotate(@degrees, @rotation) {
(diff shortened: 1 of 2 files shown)


tagliala:
Thanks for this PR.

This is going to be addressed internally in the build system, expect this change in the next release of Font Awesome

tagliala:
Fixed in 5.8.0

Thanks again!

https://github.com/FortAwesome/Font-Awesome/blob/5.8.0/scss/_mixins.scss

PR #14586, opened 2019-01-31
Fix definitions for DOM interface

danjohnso:
DOM interfaces were off slightly from published API.  Adjusted them so TS users can use accurately

I understand that:


danjohnso:
I am a typescript user and looking to use some of the SVG APIs in my project.  I noticed the definitions are off from your actual API.  I didn't review all of your interfaces, but DOM ones were quick fixes since I need them in m y project.  I am shimming my project with these definitions, but thoughts others may end up in the same boat.

tagliala:
Hi!

Thanks for this PR.

I've assigned @mlwilkerson so he can take a look

danjohnso:
Any chance this is going to get merged?

PR #14657, opened 2019-02-19
added !default flag to font-display variable

lawlesscreation:
I was looking at some of the Lighthouse performance suggestions for our application, one of which was to leverage the font-display feature, but I noticed that the font-display property for font-awesome did not have a Sass !default flag so could not be assigned.

As a side note, is the font-display feature likely to be added to FA4?


lawlesscreation:
Nevermind, there's already a PR for this 👍

PR #14825, opened 2019-03-27
Resolve typings for unit testing

favna:
This allows unit tests to be written on the library. Previously a "ts-ignore" statement would be required for this as the TypeScript Compiler would claim that "definitions" did not exist on "Library", despite the fact that when logging the full "Library" it most certianly did. I've made the property readonly since it should be immutable (as the "add" method should be used to add new definitions) and the index signature added to IconDefinitions is to satisfy the TypeScript compiler when the compiler option of "strict" (or "NoImplicitAny") is enabled.

I understand that:

### DIFF
--- a/js-packages/@fortawesome/fontawesome-common-types/index.d.ts
+++ b/js-packages/@fortawesome/fontawesome-common-types/index.d.ts
@@ -14,6 +14,14 @@ export interface IconDefinition extends IconLookup {
     string, // unicode
     string // svgPathData
   ];
+
+  [index: string]: [
+    number, // width
+    number, // height
+    string[], // ligatures
+    string, // unicode
+    string // svgPathData
+  ] | IconName | IconPrefix;
 }
 
 export interface IconPack {
--- a/js-packages/@fortawesome/fontawesome-svg-core/index.d.ts
+++ b/js-packages/@fortawesome/fontawesome-svg-core/index.d.ts
@@ -110,4 +110,5 @@ type IconDefinitionOrPack = IconDefinition | IconPack;
 export interface Library {
   add(...definitions: IconDefinitionOrPack[]): void;
   reset(): void;
+  readonly definitions: {fas: IconDefinition, fab: IconDefinition, far: IconDefinition, fal: IconDefinition};
 }


tagliala:
@robmadole @mlwilkerson could you please take a look?

Favna:
To clarify the need for this here is a sample test taken from our own code. I have gotten this to work by overriding TS Typings but that is obviously not a desired way to do things. 

```ts
import { library } from '@fortawesome/fontawesome-svg-core';
import createFontAwesomeLibrary from '../../config/fontawesome';

beforeAll(() => createFontAwesomeLibrary());

test('FontAwesome should only contain the tested icons', () => {
    const supportedIcons = Object.keys(library.definitions.fas); // This returns an array of all the icons that were added to the library in the end-user configuration
    expect(supportedIcons).toEqual(['pencil-alt', 'trash-alt']);
});
```

However without the resolved typings the compiler will throw errors:

image

For reference this is our `createFontAwesomeLibrary()`

```ts
import { library } from '@fortawesome/fontawesome-svg-core';
import { faPencilAlt, faTrashAlt } from '@fortawesome/free-solid-svg-icons';

export default () => {
  library.add(faPencilAlt, faTrashAlt);
};
```

Favna:
After some further local usage I have come to realise that local overriding (by adding a `node_modules` to the main folder, "src" f.e.) is not an interim option as such typings are not being read in the main code when doing `library.add` *sighs*.. Probably because each imported icon has its own definition file referencing the `common-types` package and so it reads those from `./node_modules` rather than the custom override one. It *might* work to set `baseURL` and `paths` in tsconfig however those values are not yet supported in Create-React-App (it's a discussion in progress) so it's not an option for my specific development environment.

That said it would be nice to see this PR merged so it can be used in our development environment without overriding the files in `./node_modules/@fortawesome`. I'm not sure how you guys think about this based on the understanding I've had to give that it would not be merged.

Favna:
@robmadole @mlwilkerson @tagliala Going by the assumption that this PR itself will not be merged as per your pull request guidelines, is there any ETA as to when the adjustments will be seen in an update to the NPM package `@fortawesome/fontawesome-svg-core`? At my company project our tests rely on these changes and for now we have the test in question set to `skip` however we would like to be able to actually run it.

mlwilkerson:
Hey @Favna, thanks for the PR. I'm looking at this now. I'll update with an ETA after I have a closer look. My initial sense is that this one can be turned around pretty quickly.

mlwilkerson:
Taking a closer look at this, it probably represents an API change, which would trigger a different level scrutiny for code review.

The [`fontawesome-svg-core` API](https://fontawesome.com/how-to-use/with-the-api/setup/getting-started) currently specifies `library.add()` but does not specify `library.definitions`. A case could be made that to access `library.definitions` directly is to work around the API to access internals.

Seen that way, this PR doesn't merely add missing typings; it suggests an expansion of the API.

So I'll still raise it for review and consideration internally, but I think this should be considered a feature request. The feature being something like: _provide access to all icon definitions known to the library_. But it's not even clear that `library.definitions` would give that, because--internally--that's not the only data structure that the library might consult to find icons. So if _provide access to all icon definitions known to the library_ were the feature, then this implementation proposed in this PR actually would not always provide the correct result.

In the meantime, please consider whether it would satisfy your testing needs to assert that `faPencilAlt` and `faTrashAlt` are found in the library. If that would suffice, then our API already supports that (if you specify the prefix). Maybe something like this:

```javascript
import { findIconDefinition, library } from '@fortawesome/fontawesome-svg-core'
import { faPencilAlt, faTrashAlt, faCoffee } from '@fortawesome/free-solid-svg-icons'

beforeAll(() => {
  library.add(faPencilAlt, faTrashAlt)
})

describe('findIconDefinition', () => {
  it('finds known icons', () => {
    expect(findIconDefinition(faPencilAlt)).toEqual(faPencilAlt)
    expect(findIconDefinition(faTrashAlt)).toEqual(faTrashAlt)
  })

  it('does not find a sample unknown icon', () => {
    expect(findIconDefinition(faCoffee)).toBeUndefined()
  })
})
```

`findIconDefinition({ iconName: 'pencil-alt', prefix: 'fas'})` would also work as a syntax for looking up a known icon by name and style prefix.

Favna:
Thank you for the information. I think the findIconDefinition will satisfy the testing needs at least for the time being so I'll be sure to implement that. From my side I'll just play the waiting game for your internal discussion then and I suppose I'll see what you guys decide in due course.

Favna:
I'd like to update this PR for potential external viewers as I have recently found a workaround that's IMHO clean enough and works in Create-React-App (which is quite strict on its tsconfig). Here goes:

First of all I learned how to add custom types for third party libs from [this article](https://www.detroitlabs.com/blog/2018/02/28/adding-custom-type-definitions-to-a-third-party-library/) which formed the basis for my new workaround. It covers the basics that will also be included below

1. Create a folder called `typings` in the root of your project
2. In that folder create a subfolder named `@fortawesome`
3. In that subfolder create a file named `index.d.ts`
4. Populate that file with the following content (comments can be removed):

```ts
// typings/@fortawesome/index.d.ts

import { IconLookup, IconName, IconPrefix } from '@fortawesome/fontawesome-svg-core';

declare module '@fortawesome/fontawesome-svg-core' {
    interface IconDefinition extends IconLookup {
        icon: [
            number, // width
            number, // height
            string[], // ligatures
            string, // unicode
            string // svgPathData
        ];

        [index: string]: [
            number, // width
            number, // height
            string[], // ligatures
            string, // unicode
            string // svgPathData
        ] | IconName | IconPrefix;
    }
    
    interface Library {
        readonly definitions: { fas: IconDefinition, fab: IconDefinition, far: IconDefinition, fal: IconDefinition };
    }
}
```
5. In your `tsconfig.json` define your new `typeRoots` and also add an exclusion to your `typings` and `node_modules` folders. The order of `typeRoots` is very important! The TypeScript compiler will first look for typings in the 0th entry, then in the 1st, in the 2nd and so on.
```json5
// tsconfig.json
{
  "compilerOptions": {
	//... other config
    "typeRoots": [ "typings", "./node_modules/@types"]
  },
  "include": [ "src" ],
  "exclude": ["node_modules", "typings"]
}
```

Now main code and matching test can be written like so:

```ts
// src/fontawesome.ts

import { library } from '@fortawesome/fontawesome-svg-core';
import { faAddressBook } from '@fortawesome/free-regular-svg-icons';
import { faExclamationTriangle, faPencilAlt, faTrashAlt } from '@fortawesome/free-solid-svg-icons';

export default () => {
  library.add(faPencilAlt, faTrashAlt, faExclamationTriangle, faAddressBook);
};
```

```ts
// src/fontawesome.test.ts

import { library } from '@fortawesome/fontawesome-svg-core';
import createFontAwesomeLibrary from './fontawesome';

beforeAll(() => createFontAwesomeLibrary());

test('FontAwesome should only contain the tested icons', () => {
  const supportedSolidIcons = Object.keys(library.definitions.fas);
  const supportedRegularIcons = Object.keys(library.definitions.far);
  expect(supportedSolidIcons).toStrictEqual(['pencil-alt', 'trash-alt', 'exclamation-triangle']);
  expect(supportedRegularIcons).toStrictEqual(['address-book']);
});
```

daviesgeek:
Is there any update on this? I'd love to be able to iterate over all the icons via `library.definitions` inside a Storybook story. I see that this would be a feature request; is there any movement towards this?

favna:
Seeing as the maintainers seem to not care about merging contributions like this and I'm cleaning up any open PRs on my name, this will be closed. I also stopped using FontAwesome in favour of either Hero Icons or Material Icons anyway.

OxyTJ:
> Is there any update on this? I'd love to be able to iterate over all the icons via `library.definitions` inside a Storybook story. I see that this would be a feature request; is there any movement towards this?

Another bump on this, I have the same use case. I've created an `IconButton` component imported into **Storybook** where I'm dynamically creating options for all of the library definitions. Any movement?

PR #15295, opened 2019-07-27
Removed Zero (0) Only

chandanndeep:
No need to write the 0 if we are using the unit with points.

### DIFF
--- a/css/all.css
+++ b/css/all.css
@@ -17,7 +17,7 @@
 
 .fa-lg {
   font-size: 1.33333em;
-  line-height: 0.75em;
+  line-height: .75em;
   vertical-align: -.0667em; }
 
 .fa-xs {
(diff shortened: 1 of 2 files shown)


tagliala:
Hi!

I just want to confirm that leading 0s are correctly stripped out in the minimized css file.

We could definitely add some lints to the normal css

chandanndeep:
@tagliala 
Take your time.

chandanndeep:
@tagliala 
Any update?

talbs:
Hi, @chandanndeep. Apologies for the dust that's settled on this PR. 

I can confirm that [our minimized CSS files strip out](https://github.com/FortAwesome/Font-Awesome/blob/master/css/all.min.css) these unnccessary `0`s. I'll be linting our CSS as part of our touchups and revamps with Font Awesome 6 shortly.

chandanndeep:
No Problem.

I checked. Now issue is fixed. So am closing the topic.

Thank You,
Chandan

PR #15639, opened 2019-10-02
Font Awesome Documention descibes ability to pass parameters to the d…

rwilson504:
Font Awesome document states that the dom.watch() interface allows for parameters to be passed to limit the scope of the watch yes these paramereters are not defined in the interface. 
 https://fontawesome.com/how-to-use/with-the-api/methods/dom-watch

I understand that:


tagliala:
Hi!

Let's assign @robmadole and @mlwilkerson

jogibear9988:
in my pull request #19301 I've made both parameters optional. Don't know if this is allowed to pass only one?

PR #15674, opened 2019-10-10
update brand.js

iamayushdas:
I understand that:


tagliala:
Hi,

thanks for this PR.

This file is being generated by our build system, could you please provide the differences between the two approaches?

iamayushdas:
> 
> 
> Hi,
> 
> thanks for this PR.
> 
> This file is being generated by our build system, could you please provide the differences between the two approaches?

I have used arrow function in my edition which was not there earlier.

BlitZz96:
This will break the whole file.
The correct IIFE with arrow function would be:
```
(() => {
  /* */
})()
```

PR #15675, opened 2019-10-10
update all.css

iamayushdas:
I understand that:


tagliala:
Hi,

thanks for this PR.

If you need a minified version, check `all.min.cs`

PR #15677, opened 2019-10-11
Update CODE_OF_CONDUCT.md

RahulSharma099:
I understand that:

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -19,6 +19,7 @@ include:
 * Gracefully accepting constructive criticism
 * Focusing on what is best for the community
 * Showing empathy towards other community members
+* Welcoming devlopers to gain or devote knowledge.
 
 Examples of unacceptable behavior by participants include:
 
@@ -28,6 +29,7 @@ advances
 * Public or private harassment
 * Publishing others' private information, such as a physical or electronic
   address, without explicit permission
+* use of unacceptable commit text or username is also prohibted.
 * Other conduct which could reasonably be considered inappropriate in a
   professional setting
 


comradekingu:
How did you envision writing "unacceptable" would be understood?

jtojnar:
This project is using Contributor Covenant code of conduct. It would be best to discuss changes upstream: https://github.com/ContributorCovenant/contributor_covenant

comradekingu:
There is a spelling error in here, in what is insult added to injury. Get rid of the CoC.

devdev606 (approved):

comradekingu:
@jtojnar The premise of adding an invariant section saying "unacceptable" is the same as any other time it is done, only now worse as a result of diverging from the original. How well do you think an appeal to logic is going to go over at the Covenant CoC repo?

mantism:
As a newcomer to this repo, why is this PR still open? shouldn't it just get closed/rejected?

comradekingu:
@mantism Maybe someone actually read the original that this breaks the licensing terms of, and understood it in terms of my critique.

There is nothing wrong with rules, so why this has to be called a CoC is beyond me.
Is conduct being codified for any particular reason?
If no pledge of allegiance is made, why have a code?

Harassment is of course illegal already, and there are rules nobody seems to read on GitHub already.
This is the only real problem:

>Other conduct which could reasonably be considered inappropriate in a professional setting

This is professionalism over community, and there the authoritarian nature of CoCs shows its true colours.
Attempts to prohibit unacceptable text in a document like that should come as no surprise to anyone.

If someone can reasonably assume what is expected in a professional environment, how is that in any way different when it is in writing from reasonably expecting what to get reprimanded or banned for when it is not?
There is no incapable person helped, and there is no ill will avoided, only a blueprint to work from and exploit if one wants to do harm. Those escalations are usually more explicit, and it is what makes a CoC different from rules.

Strip away more and it would be a mantra. Bloat isn't good, and does nothing better.

PR #15690, opened 2019-10-15
Add !important font weights, fixes #15689

### LINKED ISSUE #15689: Font weight issue when using with some frameworks
Initial bug in quasar framework
https://github.com/quasarframework/quasar/issues/5191

In that case it can be fixed with !important modifier for font-weight in .fas class (and other generic classes).

FragsterAt:
I understand that:

- [+] I'm submitting this PR for reference only. It shows an example of what I'd like to see changed but

### DIFF
--- a/css/all.css
+++ b/css/all.css
@@ -4391,7 +4391,7 @@ readers do not read off random characters that represent icons */
 @font-face {
   font-family: 'Font Awesome 5 Brands';
   font-style: normal;
-  font-weight: normal;
+  font-weight: normal !important;
   font-display: auto;
   src: url("../webfonts/fa-brands-400.eot");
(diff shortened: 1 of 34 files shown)


tagliala:
Hi!

I personally do not think that we should add `!important` property on our css rules

@robmadole any thoughts?

FragsterAt:
> I personally do not think that we should add `!important` property on our css rules
> @robmadole any thoughts?
Please, look at fiddle: https://jsfiddle.net/78Lybkhs/2/

robmadole:
Hey there. Yep, we will not add this to FA. It could have unforeseen consequences and is not the appropriate way to solve this with other frameworks. I would file an issue with Quasar and work it from that angle @FragsterAt

PR #15760, opened 2019-10-31
Release 5.11.2

nm7783:
I understand that:

  I understand that it will not be merged and I will not be listed as a contributor on this project
.<html>
</html>


kad-diable:
desconnect

Issue #8945, opened 2016-04-11
Icons are misplaced in contexts with non-zero text-indent values


crisperdue:
Using Font Awesome version 4.5.0 and Chrome 49.0.2623.110 (64-bit).

FA uses display: inline-block for icons, but this inherits text-indent values from enclosing tags.

See http://jsfiddle.net/crisperdue/1o0bbhgc/4/ for illustration and a workaround.

My suggestion is to include "text-indent: 0" in the definition of the "fa" class to positively set the text-indent for icons to 0.

crisperdue:
See http://jsfiddle.net/1o0bbhgc/ for updated version of illustration.

tagliala:
Added to 5.0.0 milestone and assigned to Dave. Thanks for reporting this

tagliala:
This is still an issue on FA 5.0.1:

https://jsfiddle.net/tagliala/tgzouajm/22/

domind:
Believe this PR should close this issue

sensibleworld:
Doing a little management in Github and (batch) closing out potentially stale issues. If the above still needs attention, please let us know and we’ll re-open!

PR #15914, opened 2019-12-06
fix for bug: #8945 Icons are misplaced with non-zero text-indent

domind:
fix for bug: #8945 Icons are misplaced in contexts with non-zero text-indent values


robmadole:
Cleaning up old PRs. Closing now.

PR #16090, opened 2020-01-19
Update TS definition to include IconDefinition as an option to IconProp

ngbrown:
At least in `@fortawesome/react-fontawesome`, the `icon` prop can take an imported icon definition.  But this is currently showing as an error in my TypeScript project because the the types don't include that.

This doesn't quite solve it, because the inferred type from the icon import doesn't seem to exactly match, but this can be taken care of with `as IconDefinition`:

```jsx
<FontAwesomeIcon icon={faExclamationCircle as IconDefinition} />
```

I understand that:


ngbrown:
Nevermind, I needed to update to a new version of the icons that included the `index.d.ts` file.

PR #17326, opened 2020-10-28
Fix usage for decorated class properties

remcohaszing:
Given the following `tsconfig.json`:

```json
{
  "compilerOptions": {
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true
  }
}
```

The following will currently causes an import error:

```ts
import { IconName } from '@fortawesome/fontawesome-common-types';

class MyClass {
  @MyDecorator
  icon: IconName;
}
```

This is because in some situations TypeScript doesn’t know that `IconName` is a
only exists as a type. It will generate an import, which can’t be resolved.

Simply adding an empty JavaScript file, will make the environment able to import
the package, even though the import value of `IconName` will be `undefined`.

I understand that:


tagliala:
Hi!

@robmadole @mlwilkerson could you please take a look?

PR #18159, opened 2021-08-11
TS support for custom icons

hanselabreu:
I understand that:

Since this library allows custom icons, we should allow accepting string as the `prefix` and `iconName`.


drusellers:
I just ran into this trying to use my custom uploaded icons. Thankfully I could use `fak` but for the icon name I had to use an existing one so TS would be ok. I would def like to see some love for custom icons in the library.

PR #18200, opened 2021-08-24
Optional callback for i2svg

cooljeffro:
I understand that:


jasonlundien:
A fix was put into 6.0.0-beta-2.

PR #18402, opened 2021-10-30
Fixed slash as a division SASS [Fixed in FA6 - our fix does not break libsass - Will be fixed in FA5 (No ETA, sorry)]

Okipa:
This PR fixes the slash-as-a-division SASS coming breaking change.

To re-execute these changes from your side, you just have to follow these simple recommandations : https://sass-lang.com/documentation/breaking-changes/slash-div

Thanks to you for applying this !

PR #18471, opened 2021-11-27
Fix deprecation warnings with the division statements

coltongit:
The SCSS libraries are now reporting deprecation warnings as of the latest version of Font Awesome (5.15.4 as of this PR). The warnings are fixed in this particular PR and can be added manually to upstream if you folks desire.

I understand that:

### DIFF
--- a/scss/_larger.scss
+++ b/scss/_larger.scss
@@ -3,8 +3,8 @@
 
 // makes the font 33% larger relative to the icon container
 .#{$fa-css-prefix}-lg {
-  font-size: (4em / 3);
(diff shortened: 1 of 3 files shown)


tagliala:
Hi!

Thanks for being part of the Font Awesome Community and thanks for this PR

This issue has been already fixed in the upstream repository for Font Awesome 6 and the fix will be backported to FA5. No ETA yet

Please also notice that our approach will preserve compatibility with libsass, while this PR only supports dart Sass

I've reopened #18402 for tracking purposes

Closing here

PR #18540, opened 2021-12-27
Use calc() for division in scss

simPod:
https://sass-lang.com/d/slash-div

Not supported since Dart Sass 2.0.0

I understand that:


tagliala:
Hi!

The issue has been already fixed in FA6 beta 3 and the fix will be backported to FA5 as soon as possible (no ETA yet, sorry)

Ref: #18402

PR #18665, opened 2022-02-09
Thumbtack.js export name regression v5 -> v6

### LINKED ISSUE #18661: v6 pro faThumbtack not working
**Describe the bug**
after upgrading to @fortawesome 6.0.0, the faThumbtack icon is broken

**To Reproduce**
Not sure how to create a reproducible example with the "pro" icons, but the issue is pretty clear from looking at the code in @fortawesome/pro-solid-svg-icons/faThumback.js. It is `require`ing itself and then exporting as `faThumbTack` instead of `faThumbtack`.

@fortawesome/pro-solid-svg-icons/faThumback.js:
```javascript
'use strict';
Object.defineProperty(exports, '__esModule', { value: true });
var source = require('./faThumbtack'); // <-- requiring self??
exports.definition = {
  prefix: source.prefix,
  iconName: source.iconName,
  icon: [
    source.width,
    source.height,
    source.aliases,
    source.unicode,
    source.svgPathData
  ]};

exports.faThumbTack = exports.definition; // <-- should be faThumbtack
exports.prefix = source.prefix;
exports.iconName = source.iconName;
exports.width = source.width;
exports.height = source.height;
exports.ligatures = source.aliases;
exports.unicode = source.unicode;
exports.svgPathData = source.svgPathData;
exports.aliases = source.aliases;
```

**Expected behavior**
faThumbtack should work like all other icons.

**Version and implementation**
Version: @fortawesome/pro-solid-svg-icons 6.0.0

- [x] SVG with JS
- [ ] Web Fonts with CSS
- [ ] SVG Sprites
- [ ] On the Desktop

**Bug report checklist**
- [x] I have filled out as much of the above information as I can
- [ ] I have included a test case because my odds go _way_ up that the team can fix this when I do
- [x] I have [searched for existing issues](https://github.com/FortAwesome/Font-Awesome/issues) and to the best of my knowledge this is not a duplicate

1st8:
Comparing
https://unpkg.com/@fortawesome/free-solid-svg-icons@6.0.0/faThumbtack.js
and
https://unpkg.com/@fortawesome/free-solid-svg-icons@5.15.4/faThumbtack.js
one can see that the export name was changed from "Thumbtack" to "ThumbTack".

This unfortunately makes the icon not importable with the babel macro.

The correct export name is "Thumbtack" as in v5.

Same issue in the pro package, which I am looking at right now.

Cheers!

I understand that:

### DIFF
--- a/js-packages/@fortawesome/free-solid-svg-icons/faThumbtack.js
+++ b/js-packages/@fortawesome/free-solid-svg-icons/faThumbtack.js
@@ -12,7 +12,7 @@ exports.definition = {
     source.svgPathData
   ]};
 
-exports.faThumbTack = exports.definition;
+exports.faThumbtack = exports.definition;
 exports.prefix = source.prefix;
 exports.iconName = source.iconName;
 exports.width = source.width;


js-packages/@fortawesome/free-solid-svg-icons/faThumbtack.js
```diff
@@ -12,12 +12,12 @@ exports.definition = {
     source.svgPathData
```
Alexei34, line 12: 1st8:patch-1

1st8:
#18661 is the better issue to track this, my patch wont fix it.

Issue #18711, opened 2022-02-16
Installing Free version with react, getting a license error (should default to free)


dk8996:
I followed the documentation on this site (https://fontawesome.com/v6/docs/web/use-with/react/). I'm getting the following error after 
```
fortawesome/fontawesome-svg-core/import.macro: config license must be either 'free' or 'pro' Learn more: https://www.npmjs.com/package/@fortawesome/fontawesome-svg-core
```
As per documentation,

I add the following package.json (npm i)
```
    "@fortawesome/fontawesome-svg-core": "^1.3.0",
    "@fortawesome/free-solid-svg-icons": "^6.0.0",
    "@fortawesome/free-regular-svg-icons": "^6.0.0",
    "@fortawesome/react-fontawesome": "latest",

```

Imports
```
import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
import { solid, regular, brands } from '@fortawesome/fontawesome-svg-core/import.macro' // <-- import styles to be used
```

dk8996:
I fixed this by adding the following to `package.json`. This should be part of the documentation.
```
  "babelMacros" : {
    "fontawesome-svg-core" : {
      "license": "free"
    }
  }
```

tagliala:
Hi!

Thanks for being part of the Font Awesome Community and apologies for the late reply.

> This should be part of the documentation.

It is actually part of the documentation: https://fontawesome.com/v6/docs/web/use-with/react/add-icons

![image](https://user-images.githubusercontent.com/556268/154558101-f8f01623-f041-434f-82a3-fb3a28236369.png)

It is not part of the setup because there are different use cases and you may not need `babelMacros`

Closing here, but feel free to comment

ghost:
@dk8996 You save me man! No need for babel-config.js and babel-plugin-macros.js when using React js and Vite. Thank you!!!

PR #18714, opened 2022-02-17
Fixing an issue where if, `babelMacros` is not set in `package.json`

dk8996:
Fixing an issue where if, `babelMacros` is not set in `package.json`, then there is an error `"config license must be either 'free' or 'pro'"` because `config = {}`. 

We need to check for the `license` (`config.license`). See my [bug report here] (https://github.com/FortAwesome/Font-Awesome/issues/18711)

I understand that:

 I'm submitting this PR for reference only. It shows an example of what I'd like to see changed but I understand that it will not be merged and I will not be listed as a contributor on this project.

### DIFF
--- a/js-packages/@fortawesome/fontawesome-svg-core/import.macro.js
+++ b/js-packages/@fortawesome/fontawesome-svg-core/import.macro.js
@@ -15,7 +15,7 @@ const styles = [
 ]
 
 function importer ({references, state, babel, source, config}) {
-  const license = (config !== undefined ? config.license : 'free')
+  const license = ((config !== undefined && config.license !== undefined) ? config.license : 'free')
 
   if (!['free', 'pro'].includes(license)) {
     throw new Error(


tagliala:
Thanks for this PR. As mentioned in the issue thread, this is documented, but having a default license set to `free` could be a good option. Anyway, for some use cases, this will require the developer to add babel-macros and keep reading the documentation 

Let's ask @robmadole

dk8996:
> Thanks for this PR. As mentioned in the issue thread, this is documented, but having a default license set to `free` could be a good option. Anyway, for some use cases, this will require the developer to add babel-macros and keep reading the documentation
> 
> Let's ask @robmadole

It seems like the code intention is to default to `free`. Also if it's required setting it should be in the set-up section, it's currently hidden in `add-icons` section. Just FYI, the search bar in your docs doesn't work.

tagliala:
> Also if it's required setting it should be in the set-up section, it's currently hidden in add-icons section. 

As per https://fontawesome.com/v6/docs/web/use-with/react/add-icons#alternative-ways-to-add-icons, having `babel-macros` is not mandatory and it depends on how developers wants to import icons, so I think that setup may not be the proper place

> Just FYI, the search bar in your docs doesn't work.

Confirmed, thanks. Will track this in a separate issue

robmadole:
Thanks @dk8996 for the report.

I think it's important enough to explicitly set `free` or `pro` in that macro config. If we don't draw attention to that I can see Pro folks missing it and then sending a ticket into support.

So for now we'll keep it.

PR #18737, opened 2022-02-22
Fix undefined operation from fa-divide in sass variable

woahitsjc:
This should fix the build when using with sass.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1,6 +1,8 @@
 // variables
 // --------------------------
 
+@import 'functions';
+
 $fa-css-prefix          : fa !default;
 $fa-style               : 900 !default;
 $fa-style-family        : "Font Awesome 6 Free" !default;


subhan402 (approved):

tagliala:
Thanks

So `fa-divide` is used in the variables partial, but the function partial is not being explicitly imported and this may lead to some issues in sass environments where functions is not being imported before variables

I think that the current implementation is correct in FA6, and the developer should import `functions` before `variables`, but would be a problem in FA5, since it will not be backward compatible for some use cases. Let's ask @robmadole and @talbs some thoughts

robmadole:
Hi @woahitsjc can you tell me how you ran into this? Do you have custom SCSS that is using `fa-divide()`?

woahitsjc:
> Hi @woahitsjc can you tell me how you ran into this? Do you have custom SCSS that is using `fa-divide()`?

So when running the build in my project, I am getting:

```
Error: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: Undefined operation "2em * fa-divide(5, 4)".
   ╷
37 │ $fa-li-margin           : $fa-li-width * fa-divide(5, 4) !default;
```

This is an issue when upgrading from FA5 to FA6.

woahitsjc:
@tagliala Did what you suggested and works. Closing this, thank you!

PR #19471, opened 2022-12-13
Corrected small spelling mistake

nocdn:
"...we are marking version 5 as Long Term Suport..."   -->   "...we are marking version 5 as Long Term Support..."

I understand that:


tagliala:
Thanks, duplicate of #19467

PR #19544, opened 2023-02-02
Fix #19423 by defaulting to 0 instead of none

### LINKED ISSUE #19423: Bug report: font-awesome script causes error in W3C HTML validator
Tested using Google Chrome on Mac OS Monterey.
Script tag in head of HTML is `<script src="https://kit.fontawesome.com/8e1ff888c3.js" crossorigin="anonymous"></script>`.
The error flagged in the validator is **Error: CSS: transform: var(--fa-rotate-angle, none) is not a transform value**.

barryp:
"transform: rotate(0);" makes sense as a default if no --fa-rotate-angle is defined.

I understand that:

### DIFF
--- a/less/_rotated-flipped.less
+++ b/less/_rotated-flipped.less
@@ -27,5 +27,5 @@
 }
 
 .@{fa-css-prefix}-rotate-by {
-  transform: rotate(~'var(--@{fa-css-prefix}-rotate-angle, none)');
+  transform: rotate(~'var(--@{fa-css-prefix}-rotate-angle, 0)');
 }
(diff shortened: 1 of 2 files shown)


tagliala:
Hi!

Thanks for being part of the Font Awesome Community and for this PR

I think this is a legit bug, because we should go either for `transform: none` or `transform: rotate(0)`, because `transform: rotate(none)` is not valid

For this specific use case, I think that `transform: rotate(0)` is the correct way to go.

/cc @robmadole

tagliala (approved):

Rushaway (approved):

Issue #19577, opened 2023-02-24
Feature request: postinstall script forces package to become unplugged with yarn pnp


mcenkar:
### Feature description

Hi,

because of [postinstall script](https://github.com/FortAwesome/Font-Awesome/blob/6.3.0/js-packages/@fortawesome/fontawesome-common-types/package.json#L29) when using `yarn` with `pnp` mode it unnecessarily unplugs package extracting all the icons from an archive. For most packages that is not a big overhead, but with FA we have thousands of files per package that are extracted.

Since this [postinstall script](https://github.com/FortAwesome/Font-Awesome/blob/6.3.0/js-packages/%40fortawesome/fontawesome-free/attribution.js) only prints licence and copyright can you consider removing it from all packages? There is possible workaround that helps but IMHO worth to consider fixing underlying issue:

```json
  "dependenciesMeta": {
    "@fortawesome/fontawesome-common-types": {
      "built": false
    },
    "@fortawesome/free-brands-svg-icons": {
      "built": false
    },
    ...
  }
```

### Feature request checklist

- [X] The title starts with "Feature request: " and is followed by the requested feature description
- [X] This is a single feature
- [X] I have [searched for existing issues](https://github.com/FortAwesome/Font-Awesome/issues) and to the best of my knowledge this is not a duplicate

mcenkar:
Just noticed there is already existing issue #18803 with much better description.

However it's been open for a while and not fixed yet so I thought I'll open PR #19578

Please let me know if I should rebase it against master branch.

PR #19578, opened 2023-02-24
fix: #19577 - remove postinstall script

mcenkar:
I understand that:


Yoviirisky (approved):

nhockool1002 (approved):
OK

tagliala:
I think this has been fixed in 6.6.0, can you please check?

PR #19815, opened 2023-07-14
Add icon list

the-one-with-raspberry:
I understand that:


tagliala:
Please use https://github.com/FortAwesome/Font-Awesome/blob/6.x/metadata/icons.json and transform it according to your needs

Please also notice that there are several styles for icons, and not all regular icons are available in FA Free, so a flat array only works for solid icons

the-one-with-raspberry:
@tagliala oh that exists? well i just wasted ~5 mins of my life doing that with node

PR #19927, opened 2023-10-03
Fix validation error in _rotated-flipped.scss

alvolex:
HTML/CSS validators give an error due to the fact that "none" was used in a css transform, changing non to 0deg fixes this issue.

Example from the Nu html checker:

Error: CSS: transform: var(--fa-rotate-angle, none) is not a transform value.

I understand that:

- [X ] I'm submitting this PR for reference only. It shows an example of what I'd like to see changed but


tagliala:
Thanks, Duplicate of #19544

PR #19963, opened 2023-10-20
fix trailing whitesaces and newlines at end of files

NicolasCARPi:
Hello,

While integrating the latest version in my app I ran into scss linting issues, so here is the change that I'd like to see in your awesome library! :)

Fixes this:
2023-10-19-190242_726x292_scrot

With best regards,
~Nico

### DIFF
--- a/scss/_animated.scss
+++ b/scss/_animated.scss
@@ -150,4 +150,3 @@
   0% { transform: rotate(0deg); }
   100% { transform: rotate(360deg); }
 }
-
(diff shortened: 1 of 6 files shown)


tagliala:
Thanks. We can't merge this directly (scss files are being generated from the build system upstream)

@robmadole some fixes are needed in the template files upstrem.

Isn't there a stylelint check of the output scss file? It should be easy to add a CI workflow for that

tagliala:
Some of those whitespaces will be fixed in the upcoming release

robmadole:
We've got all but a couple of blank lines at the end of the files fixed up in our build system. Thanks for letting us know.

NicolasCARPi:
Great!

PR #20432, opened 2024-11-22
removed unnecessary 'and' from README.md for better clarity text

Aryan2727-debug:
I understand that:


robmadole:
Thanks @Aryan2727-debug

PR #20483, opened 2024-11-27
Make typing intent clearer

bobvandevijver:
The current definition is hard to read due to the union type and with certain tooling not parsed correctly.

image

The change proposed here would solve both.

----

An alternative would be the following, but personally I prefer the `Array<>` one.

```
add(...definitions: (IconDefinitionOrPack | IconDefinitionOrPack[])[]): void;
```

----

I understand that:


robmadole:
This was released in 6.7.2. Thanks @bobvandevijver !

PR #20876, opened 2025-12-02
Moving Requests to Discussions

sensibleworld:
In order to lessen the signal-to-noise ratio in our Icon Requests, I think we should move new requests to discussions, instead of issues. This PR sets up some initial templates for us to use. The concept itself is still WIP, so we'll likely make updates/changes as we go.


robmadole (approved):

sensibleworld:
Ack! Moved a little too quickly. Looks like these are supposed to be yml files, not markdown. Another PR incoming.

PR #21443, opened 2026-01-27
Update chevron-left.svg

vatebra:
I understand that:


tagliala:
Not sure about the reasons behind this PR

image

Closing here

PR #21493, opened 2026-02-26
Align volume-up waves with speaker base to match other volume icons

GulshanJha00:
I understand that:

I noticed the wave paths are offset by 32 units relative to the speaker base.
I tested a correction by adjusting the X coordinates of the wave sections, which resolves the alignment issue.


tagliala:
Thanks. Not sure about if this is intentional, but let's assign to @sensibleworld for review

PR #21525, opened 2026-03-30
docs: fix typos in documentation

ankitkumar572005:
Fixed typographical errors across documentation markdown files using codespell.


js-packages/@fortawesome/fontawesome-svg-core/import.macro.js
```diff
@@ -67,7 +67,7 @@ function resolveReplacement({ nodePath, babel, state, macroName }) {
   }
 }
 
-// The macros corresonding to legacy style names: solid(), regular(), light(), thin(), duotone(), brands().
+// The macros corresponding to legacy style names: solid(), regular(), light(), thin(), duotone(), brands().
```
tagliala, line 70: legit

tagliala:
Thanks for the PR. I've reviewed the changes; one of the fixes is spot on (the one I've commented), but for the others, the original code is actually correct and should be maintained.

About the legit change, please do not send a new PR because that file originates from an upstream repository, I'm tracking it separately

